PR #26, opened 2024-05-30
Visualize compilation specs with `pydot`

cisaacstern:
Allows visualization of graphs using `pydot` (and `graphviz`).

This is not implemented in the CLI yet, but from a Python interpreter we can do:

```python
import yaml
from ecoscope_workflows.visualize import write_png

with open("examples/compilation-specs/time-density.yaml") as f:
    s = yaml.safe_load(f)
write_png(s, "output.png")
```
which outputs:
image


cisaacstern:
This is a little stale now, and I think we'll be evolving the compilation spec, which will break it, so coverting to draft to indicate that I'll need to revisit it (and it won't be merged in this state).

cisaacstern:
image

PR #31, opened 2024-06-14
Data connections

### LINKED ISSUE #32: Data Connection Class
We need the concept of a data connection that can be securely stored and used within various ecoscope functions. And that can ideally be serialized to a secure keystore when functions are being executed within ecoscope tasks/workflows.

cisaacstern:
Closes #32 #33


cisaacstern:
TODO:
- [ ] Omit all clients/connections from `get-params` introspection
- [ ] Add CLI subparser for `keystore` that lets us create keystores for a spec

cisaacstern:
I have presented the proposed solution to the Pydantic community for feedback here: https://github.com/pydantic/pydantic/discussions/3205#discussioncomment-9799424

cisaacstern:
I have now a draft of an interactive local "keystore" generator (i.e. .env file). Remaining TODO:

- [ ] programmatic keystore interface, for use in web app
- [ ] think about naming of this interface
- [ ] do we actually support local .env files not named ".env"?
- [ ] fix failing tests
- [ ] docs

cisaacstern:
@Yun-Wu @atmorling @walljcg, this is ready to go IMHO.

It's obviously a bit of a hard PR to review given its size and relative lack of documentation 🤭 . But it's also an important one, so before merging if you all want to / are able to give it a quick skim, perhaps something will jump out at you that you'd like to ask a question about.

I'll leave this open for now and move on to other work, and then we can merge early next week once anyone who'd like to has given it a quick read.

To clarify though, while I'm very open to criticism on this, and the possibility that I've overlooked something, I'm also quite confident in the behavior at this point, and would feel comfortable merging without review as well, if you happen to be busy with other things!

examples/dags/time_density_params.yaml
```diff
@@ -1,9 +1,5 @@
...
-  username:   # (<class 'str'>, FieldInfo(annotation=NoneType, required=True, description='EarthRanger username'), BeforeValidator(func=functools.partial(<function from_env at 0x18cc825f0>, var_name='ER_USERNAME')))
-  password:   # (<class 'str'>, FieldInfo(annotation=NoneType, required=True, description='EarthRanger password'), BeforeValidator(func=functools.partial(<function from_env at 0x18cc825f0>, var_name='ER_PASSWORD')))
-  tcp_limit:   # (<class 'int'>, FieldInfo(annotation=NoneType, required=True, description='TCP limit for EarthRanger API requests'))
-  sub_page_size:   # (<class 'int'>, FieldInfo(annotation=NoneType, required=True, description='Sub page size for EarthRanger API requests'))
+  client:   # (<class 'ecoscope_workflows.connections.EarthRangerClientProtocol'>, BeforeValidator(func=<bound method DataConnection.client_from_named_connection of <class 'ecoscope_workflows.connections.EarthRangerConnection'>>), WithJsonSchema(json_schema={'type': 'string', 'description': 'A named EarthRanger connection.'}, mode=None))
```
Yun-Wu, line 2: How does a user define a `client` in this file? Instead of the entire class, is it possible to provide an id or type configuration?

Yun-Wu: I see the example below and it seems to be a client name. Can we change the description here to be more intuitive?

ecoscope_workflows/connections.py
```diff
@@ -0,0 +1,91 @@
...
+@runtime_checkable
+class EarthRangerClientProtocol(Protocol):
+    def get_subjectgroup_observations(
+        self,
+        subject_group_name: str,
+        include_subject_details: bool,
+        include_inactive: bool,
+        since,
+        until,
+    ) -> None: ...
```
Yun-Wu, line 69: Is this class implemented somewhere else?

Yun-Wu:
Thank you Charles! It looks great! Added some minor comments for clarification here.

examples/dags/time_density_params.yaml
cisaacstern: Yes absolutely, great feedback! I will do that tomorrow before merge. 

And yes you are correct, it is given here as the id/name of a pre-configured connection, and the class is constructed at runtime. 😄

ecoscope_workflows/connections.py
cisaacstern: Yes! This protocol represents [`ecoscope.io.EarthRangerIO`](https://github.com/wildlife-dynamics/ecoscope/blob/3b87cbae882b8524a9dc9978630d1825e46be2a1/ecoscope/io/earthranger.py#L23) 😃 

That is, it represents the surface area of that object that we currently use here.

I've defined the protocol so that we can type hint the client type without taking a hard dependency on `ecoscope`, if that makes sense?

cisaacstern:
Thanks so much for the review Yun! I really appreciate it!! Left you responses inline above. 🙏

ecoscope_workflows/connections.py
Yun-Wu: Oh my bad. I meant to ask if we have implemented `get_subjectgroup_observations` somewhere else?

atmorling:
I really like this implementation :partying_face: 
Do you see the gsm solution being tenable when we get to storing user configured data connections (and therefore potentially user credentials to those data connections) via the `ecoscope-web`?

ecoscope_workflows/connections.py
cisaacstern: Gotcha! So yes, we have this task:

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/e25ecc4560de7b433e3af6ec3c11a053cbf199e3/ecoscope_workflows/tasks/io/_earthranger.py#L28-L48

which happens to have the same name as the method on this protocol, but it would not have to. The more important point is that the client inside the body of that task has the `get_subjectgroup_observations` (because it is an instance of `ecoscope.io.EarthRangerIO`) ... does that make sense? (Totally fine if not! I might just be too deep into this to see what I'm overlooking 🐰 🕳️ 😅 )

examples/dags/time_density_params.yaml
cisaacstern: I took a closer look at this and I think it will be best if we punt until we can address it via https://github.com/wildlife-dynamics/ecoscope-workflows/issues/47. Thanks for bringing this up! Very on point feedback!

ecoscope_workflows/connections.py
Yun-Wu: Oh I see. `get_client` actually returns `EarthRangerIO` which has `get_subjectgroup_observations`. In that case, do we still need this function here in the `EarthRangerClientProtocol`

Yun-Wu (approved):
Thanks a lot Charles for this wonderful work. Looking forward to getting it merged!

ecoscope_workflows/connections.py
cisaacstern: Exactly! And yes, we do need this `get_subjectgroup_observations` method placeholder here (note it's just a signature, but there is no function body, only `...`), as this is how the protocol is defined: https://typing.readthedocs.io/en/latest/spec/protocol.html#defining-a-protocol

This way we can tell Pydantic to expect "some object that supports the `get_subjectgroup_observations` interface" without needing to use the actual `EarthRangerIO` (which would require taking a compile-time dependency on `ecoscope` core, which is what I am trying to avoid) to do so.

Does that make sense? Totally fine if not!

cisaacstern:
> I really like this implementation 🥳
Do you see the gsm solution being tenable when we get to storing user configured data connections (and therefore potentially user credentials to those data connections) via ecoscope-web and then being retrieved wherever the task is executed?

Great question @atmorling. Short answer: I'm not sure. Slightly longer answer: the GSM solution here is at least a reference for how a GCP-based backed can be plugged into this PR _eventually_. At most, it's the actual solution we'd want to use, but that part I'm not sure about.

I've opened https://github.com/wildlife-dynamics/ecoscope-workflows/issues/48 to track this further, please add to the discussion there if you have further thoughts or if I haven't captured the questions correctly there!

ecoscope_workflows/connections.py
cisaacstern: (Going to merge this momentarily but feel free to continue asking questions here as needed.)

examples/dags/time_density_params.yaml
cisaacstern: One last note here pre-merge: I've also opened https://github.com/wildlife-dynamics/ecoscope-workflows/issues/49 to track this (but as noted there, this depends on #47)

cisaacstern:
Thank you @Yun-Wu and @atmorling for the reviews!!

PR #36, opened 2024-06-21
Create ETL task to map columns

Yun-Wu:
- Add a column mapping task
- Add the task to registry
- Add unittest

https://github.com/wildlife-dynamics/ecoscope-workflows/issues/34

Issue #35, opened 2024-06-17
Task: ETL - Categorical Value Classification


walljcg:
We need an ETL task that allows a user to classify categorical values on a dataframe. If there are as many unique classes as unique categorical values then this is a simple mapping of one value to another. Otherwise if the classes are less then there needs to be some logic that specifies the mapping. E.g., if x='navy blue' or x='light blue': y='blue'. 

This function should work for either numeric or text data types. 

Parameter inputs will be: 

1. Input column name
2. a dict that provides the mapping of the values of the column to the new values. If input values do not appear in the remapping dict then then are left untouched. 
3. Output column name (if it's the same as the input name then the input column is overwritten with the re-mapped values).

PR #37, opened 2024-06-21
ETL - Color Mapping

Yun-Wu:


Yun-Wu:
[#35](https://github.com/wildlife-dynamics/ecoscope-workflows/issues/35)

cisaacstern:
@Yun-Wu and @walljcg I'm curious what you'd think about categorizing this task, as well as #36, under `tasks/preprocessing`?

Is there some sense that I'm missing in which "transformation" is different than "preprocessing?

Yun-Wu:
I assumed that `preprocessing` is to preprocess some dataframe to prepare for certain analysis. Data `transformation` is more general and can be placed before or after the analysis. Happy to hear your thought too.

cisaacstern:
> I assumed that preprocessing is to preprocess some dataframe to prepare for certain analysis. Data transformation is more general and can be placed before or after the analysis.

That makes sense! I think the word "transformation" in my mind is just quite generic and a bit overloaded (then again, so is "preprocessing" 🙃 )... so this clarification is helpful. I'm happy moving forward with this naming!

PR #45, opened 2024-06-24
Split apply combine

### LINKED ISSUE #28: POC: Split-Apply-Combine using lithops
We want to be able to perform split-apply-combine operations on (geo)pandas dataframes in a distributed / scalable way both on a single multi-core machine and in the cloud. It looks like lithops may take care of the architecture for this already (https://github.com/lithops-cloud/lithops).  But we need to research how to generalize the use of lithops across tasks may need both the local/cloud split/apply/combine paradigm.

cisaacstern:
Closes #28


cisaacstern:
@walljcg here's a start on how we can leverage lithops map reduce (as demonstrated in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/28#issuecomment-2183694749) in the context of ecoscope workflows/tasks. The core task is here:

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/6e416b6a67cdc3f999b27882d4339705ff8e0ff8/ecoscope_workflows/tasks/parallelism/_lithops.py#L8-L38

And here is an example of what it might look like compiled into a runnable workflow:

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/6e416b6a67cdc3f999b27882d4339705ff8e0ff8/examples/dags/time_density_map_reduce.script_lithops.py#L79-L104

💡 Note that this does not yet run, but is more-so one step above pseudocode. Among other things, of course we don't actually need to split-apply-combine for time density maps in this way, I'm just using that as a toy example here.

In terms of where a script like this could run, if it runs locally, it can parallelize across python processes. In the cloud, we could package this script _itself_ as a "launcher" serverless function on GCP Cloud Run, which once it gets to the lithops map-reduce section, would spawn additional Cloud Run functions. The "launcher" function (running this script) would then wait for all of the parallelized tasks to complete, gather the result, and send it back to wherever we need it (database api call to the server maybe).

Note also that this pattern (launch lithops from another cloud function) is conceptually almost identical to the [Lithops Airflow Operator](https://lithops-cloud.github.io/docs/source/airflow.html#examples).

I've also iterated a bit on the workflow YAML spec design, so we can represent the map reduce operation there something like this:

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/6e416b6a67cdc3f999b27882d4339705ff8e0ff8/examples/compilation-specs/time-density.yaml#L39-L55

To make this more legible, I'm experimenting with borrowing the GitHub Actions data structure of:

```yaml
- name: "a human readable name"
  id: |
    # a unique id. we were using the task names for this before,
    # but to support reuse of the same task within a workflow, we'll need an `id`
  uses:  # the action name in github, or for us, the task importable reference
  from: |
    # i'm not actually using this here, and i don't think it's part of GitHub Actions,
    # but i thought this might be a nice way to include the github or pypi path for
    # extension task packages, which could be dynamically installed in a venv at compile
    # time (like pre-commit). https://github.com/moradology/venvception is a nice little
    # package that does this (ephemeral venvs), that a former collaborator wrote for our
    # last project.
  with:
    # kwargs to pass to the task
```

cisaacstern:
Based on the current working example [here](https://github.com/wildlife-dynamics/ecoscope-workflows/blob/split-apply-combine/examples/dags/time_density_map_reduce.script_lithops.py), some observations/thoughts re: performance and optimization (all of the below is based on local testing; I have not run on Cloud Run yet):

- For 1000 fixes, a single direct invocation of `map_function` (turns a single group into a map widget), without any `lithops` overhead, takes about 10 seconds.
- `lithops` adds about 10 seconds of overhead, for a total 20 seconds runtime (with no attempts at optimization yet).
- Of the ~10s runtime for the direct call:
    - Pydantic adds zero meanginful time (removing validation doesn't change runtime)
    - About 4 whole seconds are for calling `calculate_time_density` itself (I have not profiled further within that call yet)
    - Even with recent improvements, importing `ecoscope` core modules can be costly. If any tasks use the same `ecoscope` sub-packages, then sharing an import cache could be meaningful (I don't know if the namespace separation of just importing in function scope prevents this, but I think it may?)... but if there is _not_ module overlap, this may not matter.
    - It _is_ possible getting even more granular with import optimizations in `ecoscope` could have a meaningful impact.
    - We should also take a look at opportunities for optimization in the algorithms (i.e. `calulate_time_density`) themselves.

atmorling:
> * About 4 whole seconds are for calling `calculate_time_density` itself (I have not profiled further within that call yet)

Just throwing this up here as a data point:
etd_range

this loop and the calls to `intersect1d` are the bulk of the time
https://github.com/wildlife-dynamics/ecoscope/blob/6e7dd1ea782c2eef07d7bf794be76e7d4a077042/ecoscope/analysis/UD/etd_range.py#L191-L198

cisaacstern:
Wow @atmorling this is a heroic insight!! If the cost is largely numpy, then it should be optimizable!

Can you share how you generated this profile image? I am still a n00b when it comes to this stuff

atmorling:
This is just the result of running `cProfile` a test script visualized via [SnakeViz ](https://jiffyclub.github.io/snakeviz/), specifically:
`python -m cProfile -o etd.prof test_etd.py`
`python -m snakeviz etd.prof`

I took another look at this and there's a sort inside `intersect1d`, and we call `intersect1d` a _lot_ (ncalls in the below image)

loop-times
(this is from the same visualization as above)

cisaacstern:
Took a stap at jit-compiling that loop, so far to no avail 😄 https://github.com/wildlife-dynamics/ecoscope/pull/193

cisaacstern:
A few quick notes on build / deploy / call commands:

```console
$ lithops git:(split-apply-combine) ✗ LITHOPS_CONFIG_FILE=../../.ecoscope-workflows-tmp/.lithops_config \
lithops runtime build -b gcp_cloudrun -f Dockerfile.cloudrun ecoscope-workflows-runtime
```
```console
$ lithops git:(split-apply-combine) ✗ LITHOPS_CONFIG_FILE=../../.ecoscope-workflows-tmp/.lithops_config \
lithops runtime deploy -b gcp_cloudrun ecoscope-workflows-runtime
```
```console
$ ecoscope-workflows git:(split-apply-combine) LITHOPS_CONFIG_FILE=.ecoscope-workflows-tmp/.lithops_config \      
ECOSCOPE_WORKFLOWS_TMP=gs://ecoscope-workflows-tmp \             
python3 examples/dags/time_density_map_reduce.script_lithops.py
```

All with config file:
```yaml
lithops:
  backend: gcp_cloudrun
  storage: gcp_storage

gcp:
  region: us-central1
  credentials_path: <path to service account json>

gcp_cloudrun:
  runtime: us.gcr.io/ecoscope-poc/ecoscope-workflows-runtime
  runtime_cpu: 2
  runtime_memory: 1000
```

cisaacstern:
As of the latest commit, I am now able to generate 104 maps concurrently on Cloud Run in roughly a minute (give or take ~5-10 seconds). This is still "slow" but _it works_ 🎉 , which is a lot more than we could say this morning!

The _very_ confusing 500 errors I was seeing from Cloud Run (referenced in the [slack thread linked here](url)), turned out to be something wildly "obvious": Cloud Run timeout was set to the default 300 seconds, **_but_** the `gunicorn` server handling requests in our container still had a default _30_ second timeout 🙀 . Bumping `gunicorn` timeout up to 300 seconds seems to have completely resolved this issue.

As for why this appeared to be connected to generating EcoMaps specifically, it seems that this had nothing to do with the EcoMaps specifically, but rather that the time at which they were being generated fell across the (prior) 30 second timeout threshold, so `gunicorn` was killing these function calls before they could complete. (Relatedly, I think the reason why bumping to `lonboard`-based EcoMap seemed to help with this issue, was that it may execute a bit faster than the prior version, thus pushing more function calls below the prior 30s threshold. 🤦 )

I will open an issue `lithops` reporting this so nobody else hopefully has to relive my confusion 😄 .

cisaacstern:
The build/deploy/call commands are the same as reported in https://github.com/wildlife-dynamics/ecoscope-workflows/pull/45#issuecomment-2197911302. The current config I'm using is:

```yaml
lithops:
  backend: gcp_cloudrun
  storage: gcp_storage
  log_level: DEBUG
  monitoring: rabbitmq

rabbitmq:
  amqp_url: amqp://<RABBITMQ_USERNAME>:<RABBITMQ_PASSWORD>@<RABBITMQ_IP>:5672

gcp:
  region: us-central1
  credentials_path: <path to service account json>
gcp_cloudrun:
  runtime: us.gcr.io/ecoscope-poc/ecoscope-workflows-runtime
  runtime_cpu: 2
  runtime_memory: 4000
  worker_processes: 1
  min_workers: 1

```

Notes on config:
- I've deployed a RabbitMQ instance to GKE and the `RABBITMQ_*` values above are for that service.
- I'm probably a little overloaded on resources in this config. A single CPU is probably fine, since I have concurrency set to `1`, and somewhere around ~2500 MB is probably okay for memory, but just being conservative on resources for now so I can focus on getting things working, and we can optimize further after that.

cisaacstern:
Here are a few execution plots for concurrently generating 104 EcoMaps on Cloud Run from the latest commit. Of note here is that somewhere roughly around ~50% of the invocations are lagging until the ~30 second mark before "function start" 🟠 (you can see this tail in the histogram as well). If we could get all invocations to begin in a single group, total execution time would be closer to 30 seconds for this batch size. I have a few ideas around config tuning that may unlock that.

image
image

cisaacstern:
> Of note here is that somewhere roughly around ~50% of the invocations are lagging until the ~30 second mark before "function start" 🟠 (you can see this tail in the histogram as well). 

So yes, this invocation lag can be greatly mitigated by setting `min_workers` to a number comparable to the number of concurrent calls we expect to make, i.e.

```yaml
gcp_cloudrun:
  ...
  min_workers: 104
```

This way, there is much less time waiting for Cloud Run autoscaling, and we get _much_ less spread in "function start" 🟠 latency, as shown in the plots below. Something interesting here is that the vast majority of invocations have completed in < 40 seconds, but our end-to-end runtime is much longer due to a few outliers:

image
image

cisaacstern:
A few next steps:
- [ ] Currently we're just mapping across views of a single widget. I'd like to add widget type / name as another key into the `groups` so that we can dispatch a single map operation to generate _all views of all widgets_, that way a single map job is 1:1 with a single dashboard. I believe this will be the easiest way to reason about these jobs, insofar as it matches our distributed workloads 1:1 with our desired output.
- [ ] We still don't have a way of representing these scripts as YAML specs, but the first item in this list should come first.

cisaacstern:
TODO:
- [ ] redo filters with https://jsonforms.io/docs/multiple-choice/#one-of

e.g.

```json
"filters": [
    {
      "title": "Animal Name",
      "key": "animal_name",
      "oneOfEnum": { 
          "type": "string",
          "oneOf": [
             {
               "const": "Ao",
               "title": "Ao"
              },
             {
               "const": "Bo",
               "title": "Bo"
             }
         ]
     }
```

cisaacstern:
From discussion with @juanlescano-ng, for compatibility with `react-jsonschema-form`, we actually want to adjust the filters section so that it looks like this:

```json
{
  "schema": {
  "type": "object",
  "properties": {
    "month": {
      "type": "string",
      "enum": [
        "January",
        "February"
      ],
      "enumNames": [
        "January",
        "February"
      ],
      "default": "January"
    },
    "animal_name": {
      "type": "string",
      "enum": [
        "Ao",
        "Bo"
      ],
      "enumNames": [
        "Ao",
        "Bo"
      ],
      "default": "Ao"
    }
  },
  "uiSchema": {
    "animal_name": {
      "ui:title": "Animal Name",
      "ui:emptyValue": "Ao",
      "ui:help": "The name of the elephant"
    },
    "month": {
      "ui:title": "Month",
      "ui:help": "The month of the year."
    }
  }
}
```

this can be demonstrated by copy-and-paste here:

https://rjsf-team.github.io/react-jsonschema-form/

cisaacstern:
Here's a demo of this in action:

https://github.com/user-attachments/assets/3c019d26-d643-4cff-b21a-737a7a201d05

Issue #39, opened 2024-06-21
SPIKE: Task: Transform - Custom function evaluation


walljcg:
How feasible is it to allow users to define their own custom functions to achieve complex ETL tasks? 

Pandas provides a [Dataframe.eval()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.eval.html#pandas.DataFrame.eval) function that lets us do various column based operations. 

Pandas also provides a [Dataframe.assign()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.assign.html#pandas.DataFrame.assign) function that accepts lambda expressions as input. How safe are these and could a user mine bitcoin or print out all system variables within the lambda? 

We win a lot of functionality if we can use either df.eval() or df.assign() without creating major security concerns when running workflows in the cloud. For example df.eval ('A' + 6) or df.eval('A'*5/9) gives us a way to do value math operations whereas df.assign(lambda x: datetime(x)) lets us do things like timezone conversions or more complicated expressions.

cisaacstern:
My take is that `.eval` is clearly highly risky and probably should just be avoided entirely, unless someone can tell me something I don't know about how to guardrail it, which theres definitely a lot I don't know!

As for `.assign`, I asked ChatGPT and I think it gave a pretty compelling response that opens up a path for development there:

**Q**: "pandas docs warn about code injection risk of pd.Dataframe.eval. does pd.Dataframe.assign (which accepts a lamba function) also have code injection risk?"


_**ChatGPT says:**_

Yes, there can be a code injection risk with `pd.DataFrame.assign` if it is used improperly, particularly when lambda functions or other expressions are used in a way that involves untrusted input. The risk arises if you allow user input to directly influence the expressions or functions passed to `assign` without proper sanitization.

The `assign` method in pandas allows you to create new columns or modify existing ones using a lambda function or other callable expressions. Here’s an example of how it can be used:

```python
import pandas as pd

df = pd.DataFrame({'A': [1, 2, 3]})
df = df.assign(B=lambda x: x['A'] * 2)
print(df)
```

In this case, the lambda function is safe because it doesn’t involve any external input. However, if the lambda function or the expressions inside `assign` are constructed using external input, there is a potential for code injection.

For example, consider this unsafe usage where user input is directly used to construct the expression:

```python
user_input = "x['A'] * 2"  # Imagine this comes from an untrusted source
df = df.assign(B=lambda x: eval(user_input))
```

[...]

PR #51, opened 2024-06-25
Task: Transformations - Assign from column attribute

cisaacstern:
Towards #39

### DIFF
--- a/ecoscope_workflows/tasks/transformation/__init__.py
+++ b/ecoscope_workflows/tasks/transformation/__init__.py
@@ -1,7 +1,9 @@
+from ._assign import assign_from_column_attribute
 from ._classification import color_map
 from ._column_mapping import map_columns
 
 __all__ = [
+    "assign_from_column_attribute",
     "map_columns",
     "color_map",
 ]
--- a/ecoscope_workflows/tasks/transformation/_assign.py
+++ b/ecoscope_workflows/tasks/transformation/_assign.py
@@ -1,0 +1,27 @@
+from operator import attrgetter
+from typing import Annotated
+
+from pydantic import Field
+
+from ecoscope_workflows.annotations import DataFrame, JsonSerializableDataFrameModel
+from ecoscope_workflows.decorators import distributed
+
+
+@distributed
+def assign_from_column_attribute(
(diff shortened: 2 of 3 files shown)


cisaacstern:
This PR follows the suggestions of ChatGPT (what a world we live in) which I thought were quite reasonable, as pasted in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/39#issuecomment-2189954068, to use `pd.DataFrame.assign` in a limited way for addition of new columns based on the attributes of other columns in the dataframe.

Opening this now because I believe I will need it to address #28 / #45 ... to explain why: in #45, I want to group a dataset by month that does not already have a "month" column. AFAICT, the existing transformations we have do not support adding the "month" column based on an existing column of `pd.Timestamp` values. This small PR adds that ability.

walljcg:
Thanks Charles! 

Have you seen the [temporal and period indexers we have here](https://github.com/wildlife-dynamics/ecoscope/blob/6e7dd1ea782c2eef07d7bf794be76e7d4a077042/ecoscope/base/utils.py#L103)?  

My assumption has been that we use these functions to create the indexes and then use DataFrame.GroupBy([groupers]) to create a Grouped object which is then passed into the split-apply_combine?

cisaacstern:
A-ha, I had not been aware of those utils.

So using the linked indexers, how would I group this dataframe by month?
```python
import pandas as pd

df = pd.DataFrame(
    {
        "recorded_at": [
            pd.Timestamp("2021-01"),
            pd.Timestamp("2021-01"),
            pd.Timestamp("2021-02"),
            pd.Timestamp("2021-02"),
            pd.Timestamp("2021-03"),
        ],
        "value": [5, 6, 7, 8, 9],
    }
)
```
with this PR, it would be:
```python
from ecoscope_workflows.tasks.transformation import assign_from_column_attribute

df_new = assign_from_column_attribute(
    df, column_name="month", dotted_attribute_name="recorded_at.dt.month"
)
df_new.groupby("month")
```

cisaacstern:
I now have a different implementation of this as part of #45 which is closer to the logic in ecoscope core, so closing. Thanks for the review!

PR #58, opened 2024-07-05
make test_toml_config_table_dumps use patched tmp_path

### LINKED ISSUE #56: `test_toml_config_table_dumps` fails when a config .toml is set in the default path
I think it just needs a `with patch()` the same as the other tests

atmorling:
closes #56


cisaacstern:
Great catch!! Thanks so much!

Issue #55, opened 2024-07-05
DataConnection config fails to load from toml when name is uppercase


atmorling:
If I have a toml with a table named "[connections.earthranger.MEP_DEV]", and try to load that with something like `EarthRangerConnection.from_named_connection("MEP_DEV").get_client()` that config fails to load because we're looking only for [lowercase table names ](https://github.com/wildlife-dynamics/ecoscope-workflows/blob/08a8e05f25c2824a987effc0effbf47f59d3d0fe/ecoscope_workflows/connections.py#L32-L35) in `from_named_connection`

Issue #49, opened 2024-06-25
In JSON Schema, give EarthRangerClient params more descriptive/accurate name and description


cisaacstern:
Dependencies:

- [ ] #47 

Per @Yun-Wu's comment in https://github.com/wildlife-dynamics/ecoscope-workflows/pull/31#discussion_r1650204041, once #47 is resolved, let's make sure that `EarthRangerClient` parameters are more descriptively named and described.

The understandable confusion in the current paradigm is that while the parameter forms describe this parameter as a "client" of type "EarthRangerClient", that in fact only describes what the _eventually materialized_ object will be by the time we get into the body of the task. From the perspective of user-passed parameters, we actually need a string name of a pre-configured EarthRanger data connection.

Issue #42, opened 2024-06-21
Task: Get_Patrol_Relocations


walljcg:
We need a task to pull relocations from a patrol type from EarthRanger. This will be a very similar task to the [get_subjectgroup_observations()](https://github.com/wildlife-dynamics/ecoscope-workflows/blob/57e47cf88212c1435e7766779c3c5806197b23c2/ecoscope_workflows/tasks/io/_earthranger.py#L36) task.

cisaacstern:
duplicate of #29, which is closed, so closing

PR #59, opened 2024-07-05
Fixes for io unit test

### LINKED ISSUE #53: Unit test failure: get_subjectgroup_observations
```/Users/yunwu/MEP/ecoscope-workflows/tests/test_tasks.py::test_consumes_and_or_produces_dataframe[get_subjectgroup_observations] failed: tf = TaskFixture(task=DistributedTask(func=<function get_subjectgroup_observations at 0x147c21e10>, arg_prevalidators={}, r...'/Users/yunwu/MEP/ecoscope-workflows/ecoscope_workflows/tasks/io/get-subjectgroup-observations.example-return.parquet')
tmp_path = PosixPath('/private/var/folders/b8/nns96jhx4j7222yjslzrhnnc0000gn/T/pytest-of-yunwu/pytest-6/test_consumes_and_or_produces_0')

    @pytest.mark.parametrize(
        "tf",
        task_fixtures.values(),
        ids=task_fixtures.keys(),
    )
    def test_consumes_and_or_produces_dataframe(
        tf: TaskFixture,
        tmp_path: Path,
    ):
        args = (
            (gpd.read_parquet(tf.example_input_dataframe_path),)
            if tf.example_input_dataframe_path
            else ()  # io tasks don't have input dataframes
        )
>       in_memory = tf.task(*args, **tf.kws)

tests/test_tasks.py:149: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ecoscope_workflows/decorators.py:123: in __call__
    else func_copy(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

client = 'MEP_DEV', subject_group_name = 'Elephants', include_inactive = True
since = datetime.datetime(2011, 1, 1, 0, 0)
until = datetime.datetime(2023, 1, 1, 0, 0)

    @distributed(tags=["io"])
    def get_subjectgroup_observations(
        client: EarthRangerClient,
        subject_group_name: Annotated[
            str, Field(description="Name of EarthRanger Subject")
        ],
        include_inactive: Annotated[
            bool,
            Field(description="Whether or not to include inactive subjects"),
        ],
        since: Annotated[datetime, Field(description="Start date")],
        until: Annotated[datetime, Field(description="End date")],
    ) -> DataFrame[SubjectGroupObservationsGDFSchema]:
        """Get observations for a subject group from EarthRanger."""
>       return client.get_subjectgroup_observations(
            subject_group_name=subject_group_name,
            include_subject_details=True,
            include_inactive=include_inactive,
            since=since,
            until=until,
        )
E       AttributeError: 'str' object has no attribute 'get_subjectgroup_observations'

ecoscope_workflows/tasks/io/_earthranger.py:42: AttributeError
```

atmorling:
Closes #53 and #55
It's possible I've misunderstood the point of `test_tasks.py` and that the _real_ fix for #53 is actually covered by #49, but I was in the area and this was a good learning exercise regardless.

### DIFF
--- a/ecoscope_workflows/connections.py
+++ b/ecoscope_workflows/connections.py
@@ -32,7 +32,7 @@ class _DataConnection(_Settings):
             pyproject_toml_table_header=(
                 "connections",
                 cls.__ecoscope_connection_type__,
-                name.lower(),
+                name,
             ),
         )
         _cls = type(
--- a/tests/test_connections.py
+++ b/tests/test_connections.py
@@ -84,16 +84,17 @@ def mock_toml_config(tmp_path: Path):
 
 def test_connection_named_from_toml(mock_toml_config: Path):
     with patch("ecoscope_workflows.config.PATH", mock_toml_config):
-        conn = EarthRangerConnection.from_named_connection("mep_dev")
-        assert conn.server == "https://mep-dev.pamdas.org"
-        assert conn.username == "user"
+        with patch.dict(os.environ, clear=True):
+            conn = EarthRangerConnection.from_named_connection("mep_dev")
+            assert conn.server == "https://mep-dev.pamdas.org"
+            assert conn.username == "user"
 
-        assert isinstance(conn.password, SecretStr)
-        assert str(conn.password) == "**********"
-        assert conn.password.get_secret_value() == "pass"
+            assert isinstance(conn.password, SecretStr)
+            assert str(conn.password) == "**********"
+            assert conn.password.get_secret_value() == "pass"
 
-        assert conn.tcp_limit == 5
-        assert conn.sub_page_size == 4000
+            assert conn.tcp_limit == 5
+            assert conn.sub_page_size == 4000
 
 
 @pytest.fixture
@@ -114,25 +115,28 @@ def mock_toml_config_no_secrets(tmp_path: Path):
 
 
 def test_connection_named_from_toml_with_env_secrets(mock_toml_config_no_secrets: Path):
-    with patch("ecoscope_workflows.config.PATH", mock_toml_config_no_secrets):
-        with pytest.raises(ValidationError, match="password"):
-            _ = EarthRangerConnection.from_named_connection("mep_dev")
+    with patch.dict(os.environ, clear=True):
+        with patch("ecoscope_workflows.config.PATH", mock_toml_config_no_secrets):
+            with pytest.raises(ValidationError, match="password"):
+                _ = EarthRangerConnection.from_named_connection("mep_dev")
 
-    with patch("ecoscope_workflows.config.PATH", mock_toml_config_no_secrets):
-        with patch.dict(
-            os.environ,
-            {"ECOSCOPE_WORKFLOWS__CONNECTIONS__EARTHRANGER__MEP_DEV__PASSWORD": "pass"},
-        ):
-            conn = EarthRangerConnection.from_named_connection("mep_dev")
-            assert conn.server == "https://mep-dev.pamdas.org"
-            assert conn.username == "user"
+        with patch("ecoscope_workflows.config.PATH", mock_toml_config_no_secrets):
+            with patch.dict(
+                os.environ,
+                {
+                    "ECOSCOPE_WORKFLOWS__CONNECTIONS__EARTHRANGER__MEP_DEV__PASSWORD": "pass"
+                },
+            ):
+                conn = EarthRangerConnection.from_named_connection("mep_dev")
+                assert conn.server == "https://mep-dev.pamdas.org"
+                assert conn.username == "user"
 
-            assert isinstance(conn.password, SecretStr)
-            assert str(conn.password) == "**********"
-            assert conn.password.get_secret_value() == "pass"
+                assert isinstance(conn.password, SecretStr)
+                assert str(conn.password) == "**********"
+                assert conn.password.get_secret_value() == "pass"
 
-            assert conn.tcp_limit == 5
-            assert conn.sub_page_size == 4000
+                assert conn.tcp_limit == 5
+                assert conn.sub_page_size == 4000
 
 
 def test_resolve_client_from_env(named_mock_env):
--- a/tests/test_tasks.py
+++ b/tests/test_tasks.py
@@ -1,5 +1,4 @@
 from dataclasses import dataclass
-from datetime import datetime
 from importlib.resources import files
 from pathlib import Path
 from typing import Callable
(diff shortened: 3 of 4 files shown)


tests/test_tasks.py
```diff
@@ -35,7 +36,9 @@ class TaskFixture:
             input_dataframe_arg_name="",
             example_input_dataframe_path="",
             kws=dict(
-                client="MEP_DEV",
+                client=EarthRangerConnection.from_named_connection(
+                    "MEP_DEV"
+                ).get_client(),
```
cisaacstern, line 41:
```suggestion
                client=EarthRangerConnection(
                    server=os.environ["EARTHRANGER_SERVER"],
                    username=os.environ["EARTHRANGER_USERNAME"],
                    password=os.environ["EARTHRANGER_PASSWORD"],
                    tcp_limit=...,
                    sub_page_size=...,
                ).get_client(),
```

ecoscope_workflows/connections.py
```diff
@@ -32,7 +32,7 @@ def from_named_connection(
             pyproject_toml_table_header=(
                 "connections",
                 cls.__ecoscope_connection_type__,
-                name.lower(),
+                name,
```
cisaacstern, line 35: What does this give us?

tests/test_connections.py
```diff
@@ -84,16 +84,17 @@ def mock_toml_config(tmp_path: Path):
 
 def test_connection_named_from_toml(mock_toml_config: Path):
     with patch("ecoscope_workflows.config.PATH", mock_toml_config):
-        conn = EarthRangerConnection.from_named_connection("mep_dev")
-        assert conn.server == "https://mep-dev.pamdas.org"
-        assert conn.username == "user"
+        with patch.dict(os.environ, clear=True):
```
cisaacstern, line 87: To understand, this is just to make sure we don't have random variables in the environment that could pollute the test?

tests/test_connections.py
```diff
@@ -114,25 +115,28 @@ def mock_toml_config_no_secrets(tmp_path: Path):
...
-            assert conn.server == "https://mep-dev.pamdas.org"
-            assert conn.username == "user"
-
-            assert isinstance(conn.password, SecretStr)
-            assert str(conn.password) == "**********"
-            assert conn.password.get_secret_value() == "pass"
-
-            assert conn.tcp_limit == 5
-            assert conn.sub_page_size == 4000
+    with patch.dict(os.environ, clear=True):
```
cisaacstern, line 118: Same question as above, this is just to make sure we don't have a polluted env?

cisaacstern (commented):
Thanks for working on this @atmorling! As we discussed on video:
- I think it makes sense to instantiate the `EarthRangerConnection` explicitly from the environment for this test, to avoid having to deal with the `.from_named_connection` machinery here
- Let's move this test into it's own module in the `tests/tasks/` directory and mark it with a [`pytestmark = pytest.mark.io`](https://docs.pytest.org/en/7.1.x/example/markers.html#marking-whole-classes-or-modules) global mark, so that we can skip it if we want.

ecoscope_workflows/connections.py
atmorling: This is to fix #55

tests/test_connections.py
atmorling: Correct, otherwise `from_named_connection` will resolve the 'stray' env vars.
This is pretty limited in scope to collisions with "mep_dev" and we could rename the connections defined in `test_connections` but since the tests are explicitly testing specific combinations of toml/env config, I think it makes sense to clear any existing env vars first. That way we never have to worry.

tests/test_connections.py
atmorling: Yep!

tests/test_tasks.py
atmorling: How do you feel about the fixture structure and the paradigm of pumping those through `test_consumes_and_or_produces_dataframe`? I'm in two minds whether to duplicate that logic or convert the ER test to a more standard looking set of tests around the task similar to `test_column_mapping` etc

ecoscope_workflows/connections.py
cisaacstern: Cool, nice fix. And the env vars are not case sensitive, so this should work regardless of toml or env var?

atmorling: Env vars are kind of case sensitive, I guess the way Pydantic/`SettingsConfigDict` loads them isn't though (otherwise this wouldn't work). 
I'll test that some more, I'd like to understand what happens when we have upper and lowercase MEP Dev env vars for example.
Might be better to remove the `.lower()` [here](https://github.com/wildlife-dynamics/ecoscope-workflows/blob/948dfb855388647180fa7c81faa2f027e08c1152/ecoscope_workflows/connections.py#L29) as well

tests/test_tasks.py
cisaacstern: @atmorling apologies for the delay on this.

I'd say let's do whatever is fastest here to unblock @Yun-Wu in #42 which I think is just leave this as a parametrization of `test_consumes_and_or_produces_dataframe`. But let's open an issue and/or remember that before long, moving this into its own test module is a good idea for better readability.

cisaacstern: That being said, if you think you can succinctly/easily split this off into its own module now, then I'd say go for it! This is where we should be headed soon, I think, as the style that Yun established with separate modules is a lot clearer than what I'd initially done here.

So if it feels easy (or close enough to easy), go for it here! If not, let's punt to unblock #42.

ecoscope_workflows/connections.py
cisaacstern: Yes I think the way we've set it up Pydantic is not case sensitive when it comes to env vars https://github.com/wildlife-dynamics/ecoscope-workflows/blob/948dfb855388647180fa7c81faa2f027e08c1152/ecoscope_workflows/connections.py#L31

cisaacstern:
Looking awesome!! Way to just make the modularization happen 😄 

I'll be curious to know if you have any suggestions of how to handle testing better than asserting against cached parquet. I'm concerned that might not scale well or perhaps be hard to maintain? Don't need to worry about that for this PR of course.

atmorling:
Yeah it's definitely not ideal, it's an issue in `ecoscope` as well obviously.
Specifically with ER, there's a bug in the API that means data for returned something like `get_subjectgroup_observations` is different _per user_, which means tests like this would never pass for me unless I regenerated the parquet file.

Something I think might work is validating the structure of the dataframe, so we only make sure that we have the columns we expect and the data in the id columns look like uuids, the times look like timestamps/epochs, etc. and don't care about things like how many rows/do the values match.
To me that seems like the right middle ground between `assert result not empty` and `assert result.matches(source_of_truth.parquet)`

cisaacstern:
I like that. And we already have the dataframe schema as part of this package 🙌 

Are you happy with this PR in its current state? If so I'll merge!

cisaacstern:
🚀

Issue #61, opened 2024-07-06
Update EcoMap task 


atmorling:
Update this task to handle the new EcoMap interface

PR #64, opened 2024-07-10
Update EcoMap task

atmorling:
Brings it in line with the new version/interface.

This doesn't allow multiple dataframe as inputs as yet, I think we eventually want a `create_map_layer` that allows separation of styling a map layer/knowledge of s specific dataframe from creating the map and its widgets


atmorling:
Closes #61

cisaacstern (approved):
Wow, @atmorling! Way to bang this one out, and also just telepathically intuit (?) how to re-generate the examples using the cli. (Maybe that's actually documented in the README, but still, I'd say non-obvious given how young this project is, even if it is documented!)

Really appreciate this. I agree about layers, and also this a great starting place!

PR #80, opened 2024-07-17
Create the simplest version of a patrol workflow

Yun-Wu:
- Create patrol_workflow spec and generate its corresponding script and params. It includes `get_patrol_observations`, `process_relocations`, `relocations_to_trajectory`, `calculate_time_density`, `draw_ecomap`
- Fix validation errors for ER data connection by removing EarthRangerClientProtocol
- Fix io unittest failure


Yun-Wu:
@cisaacstern `EarthRangerClientProtocol` is causing me validation errors as `EarthRangerIO != EarthRangerClientProtocol`. I had to remove it in the task and use `EarthRangerIO` directly. If I understand correctly, it is used as the interface between the data connection and the actual client. I removed it and it works fine. Could you please take a look at my fix and see if it makes sense to you?

cisaacstern (changes requested):
Hi @Yun-Wu apologies for the delay on this. 

> EarthRangerClientProtocol is causing me validation errors as EarthRangerIO != EarthRangerClientProtocol. I had to remove it in the task and use EarthRangerIO directly.

Sorry to hear that this was not working, clearly I/we have some work to do regarding documentation and/or developer experience here.

I am not comfortable with your fix here, however, as it introduces a "compile time" dependency on `ecoscope` core, which AFAICT we do not otherwise have.

My intention in using the Protocol as a stand-in for the actual EarthRangerIO object was to prevent taking this compile-time dependency on `ecoscope` core, and defer `ecoscope` core to being a runtime dependency only.

For an example of why this might matter, if/when we eventually run a microservice that compiles custom workflows, with `ecoscope` core imported at the module top level, we then require `ecoscope` (and, in turn, any of its dependencies) to be available to the compiler container. By using the Protocol as a stand-in at the module/outer layer, as I'm hoping we can do here, we avoid that dependency, which IMHO is beneficial. (Among other things, it means workflows compilation is only loosely and not tightly coupled to `ecosope` core dependency changes.)

Does that goal make sense?

If you have time sometime after one of the next few standups in the next few days, I'm happy to sync on debugging this. If you prefer, I could also give it a shot myself and then get back to you with any conclusions.

Yun-Wu:
@cisaacstern sure. Created this ticket to track the fix and assigned to you: https://github.com/wildlife-dynamics/ecoscope-workflows/issues/84

ecoscope_workflows/tasks/results/_ecomap.py
```diff
@@ -75,4 +78,11 @@ def draw_ecomap(
...
     m.zoom_to_bounds(m.layers)
-    return m.to_html()
+    return_result = output_path
+    if output_path:
+        m.to_html(output_path)
+    else:
+        return_result = m.to_html()
+
+    # IMPORTANT: EXPLICITLY INCLUDE A RETURN STATEMENT AS AST IS NOT ABLE TO RECOGNIZE IT WITHIN AN IF STATEMENT
+    return return_result
```
cisaacstern, line 88: Rather than saving the HTML here, let's use #89 (or whatever that becomes post-review) to persist the text as part of the workflow definition.

ecoscope_workflows/tasks/results/_ecoplot.py
```diff
@@ -49,13 +52,26 @@ def draw_ecoplot(
...
+        return plot.to_html(
+            default_height="100%",
+            default_width="100%",
+            config={
+                "autosizable": True,
+                "fillFrame": True,
+                "responsive": True,
+                "displayModeBar": False,
+            },
+        )
```
cisaacstern, line 77: Same comment as above re: using #89

examples/compilation-specs/patrol_workflow.yaml
```diff
@@ -0,0 +1,10 @@
+name: patrol_workflow
+cache_root: gcs://my-bucket/ecoscope/cache/dag-runs
+tasks:
+  get_patrol_observations: {}
+  process_relocations:
+    observations: get_patrol_observations
+  relocations_to_trajectory:
+    relocations: process_relocations
+  draw_ecomap:
+    geodataframe: relocations_to_trajectory
```
cisaacstern, line 10:
```suggestion
  draw_ecomap:
    geodataframe: relocations_to_trajectory
  persist_text:
    text: draw_ecomap
```

Here's what using the task added in #89 would look like here, but #89 gives a fuller example in the tests.

cisaacstern (commented):
Here's a quick targeted review re: persisiting results. I've just opened #89 (which closes #81). Instead of handling persistence of results in the ecomap/plot tasks, we can use that task to do so.

Lmk if that seems like it can work, or if you have any concerns. 🙏

cisaacstern (commented):
This is awesome!!! 🤩 Thanks so much!

cisaacstern (approved):

PR #89, opened 2024-07-19
[Task] Persist text result to local or remote storage

### LINKED ISSUE #81: Task: Write Result to Local Storage

cisaacstern:
Closes #81 

Adapted from #45


cisaacstern:
TODO:

- [ ] add to time density workflow

cisaacstern:
@Yun-Wu would love your perspective on this. I will make a note over in #80 as well, as I this can be useful there I believe. Please let me know if anything here is not self-explanatory or could use further explanation.

Yun-Wu (approved):

PR #90, opened 2024-07-19
Yaml spec expressiveness 

### LINKED ISSUE #50: Rework compilation spec schema to look more like GitHub Workflows
The current compilation spec schema is represented by this example:

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/08a8e05f25c2824a987effc0effbf47f59d3d0fe/examples/compilation-specs/time-density.yaml#L1-L12

This has worked so far but is not expressive enough to capture the full range of relationships we will want to encode in these specs. As first noted in https://github.com/wildlife-dynamics/ecoscope-workflows/pull/45#issuecomment-2185513124, I believe we will be better served by adapting a schema inspired by GitHub Workflows `jobs.step`, i.e.

```yaml
- name: "a human readable name"
  id: |
    # a unique id. we were using the task names for this before,
    # but to support reuse of the same task within a workflow, we'll need an `id`
  uses:  # the action name in github, or for us, the task importable reference
  from: |
    # i'm not actually using this here, and i don't think it's part of GitHub Actions,
    # but i thought this might be a nice way to include the github or pypi path for
    # extension task packages, which could be dynamically installed in a venv at compile
    # time (like pre-commit). https://github.com/moradology/venvception is a nice little
    # package that does this (ephemeral venvs), that a former collaborator wrote for our
    # last project.
  with:
    # kwargs to pass to the task
```

cisaacstern:
WIP closes #50 when complete

Opening as a draft just to get ideas out there


cisaacstern:
Noting to self that Python's [builtin `map`](https://docs.python.org/3/library/functions.html#map) can serve as a stand-in for Lithops FunctionExecutor.map, to allow addition of a `map` operator by this PR without implicating the additional complexity of Lithops, which can be handled in a follow-on to this.

cisaacstern:
A few notes on next steps:
- This can be opened for review following merge of:
    - #91
    - #94
- This does not currently actually _run_ a mapped task end-to-end, but rather puts the structure in place to be able to do that. To keep things relatively maneageable, I suggest we keep it that way, and handle the actual tested implementation in a quick follow-on PR. (As this one is large enough as it is).
- The first follow-on can just get some version of a simple parallel mapped task to _run_ (like the one demonstrated here).
- The second follow-on add a lithops template to do the same thing but with lithops (not using python builtins)
- Next, we can then do a "real" e2d workflow using map + sequential non-map tasks (i.e. getting time-density.yaml to match #45
- After that. we can introduce lithops `call_async` for the non-mapped calls in the lithops template, and figure out the right way to gather results to be as non-blocking as possible there.

Some other random thoughts:
- Lithops _does_ handle (pickle-based) serialization of arbitrary Python objects, includind dataframes. So to start with, we may be able to use pandas groupby directly (even for lithops) and then only later implement hive partitioning (as in #45) for performance comparison. (I assume it will be faster, but we will see.)

cisaacstern:
Linked issue checklist:
- [ ] #105
- [ ] #106
- [ ] #107

cisaacstern:
Re: #107, as of writing this, I have a placeholder here using a very minimal stub implementation of `split`... on further thought, and as mentioned in the revised comment above, I think it makes most sense to revert this placeholder in favor of a runnable implementation using "array inputs" as the input for a task that maps over those inputs.

So this suggests that array inputs to `with` have two possible uses:
- In `mode="call"` they allow specifying merging of iterables (the resulting array will be passed as _as an array_  (i.e. list) to a single call
- In `mode="map"` they are (one way) to specify the elements to map over
- In (an as-yet unimplemented `mode="flatmap"`) ... they can be a way to specify multiple parallel collects to flat map over (e.g. the return values of multiple _split_) calls.

README.md
```diff
@@ -95,43 +95,73 @@ def my_cool_analysis_task(
...
+  # this next task has a dependency defined by the `with` field...
+  - name: Turn Observations into Relocations
+    id: relocs
+    task: process_relocations
+    with:
+      # this k:v pair indicates that the `process_relocations` task takes an argument
+      # `observations`, and we want the value passed to this argument to be the return
+      # value of the task instance with id `obs`, which is the root task of this workflow.
+      # note that (a) variables which resolve based on the outputs of other tasks are wrapped
+      # in `${{ }}`; and (b) return values of other task instances are referenced by strings
```
walljcg, line 139: so (b) means a task has no previous dependencies?

cisaacstern: I'm not sure that I totally understand your question but I _think_ the answer is no.

`(b)` reads in full:

```python
# ... (b) return values of other task instances are referenced by strings
# having the structure `workflow.<id>.return`.
```

which is meant to indicate, for example, that we can reference the return value of a task instance defined like this:

```yaml
- name: Get SubjectGroup Observations from EarthRanger
  id: obs
  task: get_subjectgroup_observations
```

using a `with` block like this:

```yaml
with:
 observations: ${{ workflow.obs.return }}
```

README.md
```diff
@@ -95,43 +95,73 @@ def my_cool_analysis_task(
 
 Compilation specs define a DAG of known [tasks](#tasks), which can either be built-ins,
 or tasks registered via the [extensible task registry](#extensible-task-registry).
+Tasks are composed into workflows as a sequence of "task instance" objects, each of which
+declares a step of a workflow, and has the following fields:
 
-The inline comments in this example explain what each line means:
+```yaml
+- name:  # [required] a human readable name for the step
```
Yun-Wu, line 102: where do we use `name`?

examples/compilation-specs/patrol_workflow.yaml
```diff
@@ -1,21 +1,70 @@
...
+    with:
+      text: ${{ workflow.patrol_events_ecomap.return }}
+      root_path: ${{ env.ECOSCOPE_WORKFLOWS_RESULTS }}
+  - name: Create Map Widget for Patrol Events
+    id: patrol_events_map_widget
+    task: create_map_widget_single_view
+    with:
+      # FIXME: i had this typo and it was not caught by validator, i.e.
+      # we cannot be allow texts to reference their own ids as dependencies
+      # data: ${{ patrol_events_map_widget.return }}
```
Yun-Wu, line 60: Do we have a validator of this spec?

ecoscope_workflows/compiler.py
```diff
@@ -62,69 +229,82 @@ def ruff_formatted(returns_str_func: Callable[..., str]) -> Callable:
...
+                dep_list = [d for d in dep] if isinstance(dep, list) else [dep]
+                for d in dep_list:
+                    if isinstance(d, TaskIdVariable) and d.value not in all_ids:
+                        raise ValueError(
+                            f"Task `{task_instance.name}` has an arg dependency `{d.value}` that is "
+                            f"not a valid task id. Valid task ids for this workflow are: {all_ids}"
+                        )
+        return self
+
+    # TODO: on __init__ (or in cached_property), sort tasks
```
Yun-Wu, line 294: What's the current order of execution?

examples/compilation-specs/patrol_workflow.yaml
cisaacstern: Good catch! Fixed that just now in [eb0e967](https://github.com/wildlife-dynamics/ecoscope-workflows/pull/90/commits/eb0e967c561fd71e37d73cafc1d0a74a6798de65)

README.md
cisaacstern: Right now this is used in a few ways:

1. To provide descriptive error messages pointing the user to a particular task instance in their workflow, e.g.
https://github.com/wildlife-dynamics/ecoscope-workflows/blob/e1f66bf07a024d3d5d983585b1e274d736cbf79c/tests/test_compiler.py#L109-L112

cisaacstern: Your question made me realize two things I needed to fix:

- We should also be using these names for the cell headers in jupytext, fixed that just now in [4ce2ab5](https://github.com/wildlife-dynamics/ecoscope-workflows/pull/90/commits/4ce2ab514605e0f6da50442fa345bb017ec74d46)
- In the task instances, these `name`s are the human-readable identifiers, whereas the `id`s are the programmatic identifiers. Therefore I realized it was confusing that the top-level `name` of the compilation spec was (as used in this PR) in fact _not_ supposed to be a human-readable name, but rather a programmatic identifier. I therefore renamed this field to `id` in [fc97e10](https://github.com/wildlife-dynamics/ecoscope-workflows/pull/90/commits/fc97e1072432ad9d2eafea4cbb595e33b28be296)

ecoscope_workflows/compiler.py
cisaacstern: The current order of execution follows the order of the task instance list in the spec, so for a spec like

```yaml
id: calculate_time_density
workflow:
  - name: Get Subjectgroup Observations
    id: obs
    task: get_subjectgroup_observations
  - name: Process Relocations
    id: relocs
    task: process_relocations
    with:
      observations: ${{ workflow.obs.return }}
  - name: Transform Relocations to Trajectories
    id: traj
    task: relocations_to_trajectory
    with:
      relocations: ${{ workflow.relocs.return }}
```

the execution order would be:

1. Get Subjectgroup Observations
2. Process Relocations
3. Transform Relocations to Trajectories

This also applies if we are combining DAG branches (via either reduction or mapping) so the order of execution for

```
id: create_dashboard
workflow:
  - name: Create Map Widget Single View
    id: map_widget
    task: draw_ecomap
  - name: Create Plot Widget Single View
    id: plot_widget
    task: draw_ecoplot
  - name: Gather Dashboard
    id: dashboard
    task: gather_dashboard
    with:
      widgets:
        - ${{ workflow.map_widget.return }}
        - ${{ workflow.plot_widget.return }}
```

would be:

1. Create Map Widget Single View
2. Create Plot Widget Single View
3. Gather Dashboard

And for

```yaml
id: calculate_time_density
workflow:
  - name: Get Subjectgroup Observations A
    id: obs_a
    task: get_subjectgroup_observations
  - name: Get Subjectgroup Observations B
    id: obs_b
    task: get_subjectgroup_observations
  - name: Draw Ecomaps
    id: ecomaps
    task: draw_ecomap
    mode: "map"
    iter:
      geodataframe:
        - ${{ workflow.obs_a.return }}
        - ${{ workflow.obs_b.return }}

```

would be:

1. Get Subjectgroup Observations A
2. Get Subjectgroup Observations B
3. Draw Ecomaps

cisaacstern: So this will obviously create problems if all dependencies of a task are not already executed by the time it is executed (because then some of its specified arguments will be undefined).

While I was writing this response I realized that it's actually not that hard to raise a ValidationError if the list is not correctly sorted, so I added that check in [a2cc7da](https://github.com/wildlife-dynamics/ecoscope-workflows/pull/90/commits/a2cc7da3c6875c4474b95472061754199cfa5cc5) + [e30c013](https://github.com/wildlife-dynamics/ecoscope-workflows/pull/90/commits/e30c013484f83bc0122fcf7e49a84b61982feb15)

cisaacstern: This check doesn't actually sort the dag (which the comment you were asking about suggested we do) ... but the more I think about it, I realized I am not sure if we actually want to do that or not. I opened https://github.com/wildlife-dynamics/ecoscope-workflows/issues/118 for us to consider that.

cisaacstern:
Thanks all for the reviews!

Particularly @Yun-Wu whose questions prompted a quick series of last-minute fixes + tweaks + enhancements which I would not have otherwise thought of.

Going to merge to keep moving forward.

PR #91, opened 2024-07-19
Tasks for creating and merging widget views

### LINKED ISSUE #67: Task: Create Widget
Implement a task to create a widget, including data/graphs required for a dashboard. It should support the following types:
1. Single value
2. Chart
3. Ecomap
4. Static widget: number/picture

cisaacstern:
Closes #67 

Adapted from #45 

WIP so opening as draft


cisaacstern:
Switched the base to `drop-ast` (#98), so I can proceed with development without being blocked by the issues described in #95.

Once #98 goes in, I can switch the base back to `main`.

cisaacstern:
@Yun-Wu @atmorling @walljcg thanks in advance for taking a look at this.

Some of this may not make total sense outside the context of #94, so if you'd like, please feel free to take a look for reference at #94 (which is nearly done, and I will open for review after this goes in).

And even with the context of #94, it might be hard to grasp the "point" of some of this until you can see it end-to-end. I've been synchronizing work between this and #94, so feel quite confident this will all "work" in the end, but figured it's better to atomize the PRs as much as possible, so opening this as its own PR, rather than a mega-PR combined with #94.

tests/tasks/test_widgets.py
```diff
@@ -0,0 +1,318 @@
...
+    )
+    widget2 = GroupedWidget(
+        widget_type="map",
+        title="A Great Map",
+        views={
+            ("month", "=", "march"): "/path/to/precomputed/mar/map.html",
+            ("month", "=", "april"): "/path/to/precomputed/apr/map.html",
+        },
+    )
+    widget1 |= widget2
```
walljcg, line 91: I'm curious about how this in-place OR operator would work on the title / widget_type fields?

cisaacstern: It leaves them unchanged. The only thing updated is `views`:
https://github.com/wildlife-dynamics/ecoscope-workflows/blob/51a6680af3bc82f7acccb389c70bffd6ebb4def9/ecoscope_workflows/tasks/results/_widget_types.py#L65-L71

cisaacstern: As you can see from above, if merge_keys don't match, the widgets are not allowed to be `|=` in place OR'ed ... and we throw a ValueError

cisaacstern: https://github.com/wildlife-dynamics/ecoscope-workflows/blob/51a6680af3bc82f7acccb389c70bffd6ebb4def9/ecoscope_workflows/tasks/results/_widget_types.py#L58-L61

cisaacstern: And the merge_key is the sum of the title and widget_type

cisaacstern: Stated another way, the title and widget_type of `self` are preserved, but the merge is not allowed to happen if those fields are different in `other`

ecoscope_workflows/tasks/results/_widget_types.py
walljcg, line 1: should we de-couple the 'results' from the 'wdigets' in a dashboard that will display them a bit more? E.g., if we replace the name 'widget' here with 'result' instead? Because running a workflow may not necessarily always end up in a dashboard. And since we're not conveying anything about the display of the widgets/results here then it's really 'results data'. Happy to discuss this further!

ecoscope_workflows/tasks/results/_widget_types.py
```diff
@@ -0,0 +1,71 @@
...
+        """Get a WidgetSingleView for a specific view."""
+        return WidgetSingleView(
+            widget_type=self.widget_type,
+            title=self.title,
+            view=view,
+            data=self.views[view],
+        )
+
+    @property
+    def merge_key(self) -> GroupedWidgetMergeKey:
```
Yun-Wu, line 59: why do we need this?

ecoscope_workflows/tasks/results/_widget_tasks.py
```diff
@@ -0,0 +1,147 @@
...
+    GroupedWidgetMergeKey,
+    PrecomputedHTMLWidgetData,
+    SingleValueWidgetData,
+    TextWidgetData,
+    WidgetSingleView,
+)
+
+
+@distributed
+def create_map_widget_single_view(
```
Yun-Wu, line 18: do we want to configure layout information, e.g. default width/height for widgets here?

cisaacstern: At this point that would be handled by the server. There will be a paired PR to this forthcoming that adds layout info including those defaults.

ecoscope_workflows/tasks/results/_widget_types.py
cisaacstern: In the distributed setting we may be generating different views of the same widget in separate concurrent processes. When it comes time to combine those views into a single widget we can use this key to know if two widget objects are mergeable (e.g. they are views of the same widget).

ecoscope_workflows/tasks/results/_widget_tasks.py
```diff
@@ -0,0 +1,147 @@
...
+        view=view,
+        data=data,
+    )
+
+
+@distributed
+def create_plot_widget_single_view(
+    title: Annotated[str, Field(description="The title of the widget")],
+    data: Annotated[
+        PrecomputedHTMLWidgetData, Field(description="Path to precomputed HTML")
```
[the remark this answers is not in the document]
cisaacstern: At this moment, no. This is narrowly focused at serving data for the FE dashboard and IIUC the only data required for that at the moment is HTML. Happy to support other formats as needed!

ecoscope_workflows/tasks/results/_widget_types.py
cisaacstern: Interesting. In the interest of the current sprint to stably recreate #45 and get a Patrols Dashboard running by end of next week, would you be comfortable merging this as-is here, and revisiting this larger discussion as part of a new issue (with potential follow-up PR)?

I do agree this is a lot of effort here just for "Widgets" and am totally in favor of generalizing this to be more useful in other applications if at all possible.

That being said, in pursuit of the dashboard spring, I have a lot of work cantilevered out over the top of this PR (namely, #94, and #90 on top of that), and I'd strongly prefer not making large structural changes here until all of that goes in, if possible.

cisaacstern: That being said, if this is truly just a matter of renaming widget -> result and there's a clear reason for that, then maybe we can include it here. Let's discuss tomorrow (Friday).

walljcg: Thanks Charles. Yep very happy for you to merge and don't want to cause any additional work! Let's chat it through after standup to see if it makes sense.

cisaacstern: opened https://github.com/wildlife-dynamics/ecoscope-workflows/issues/101 to track this

ecoscope_workflows/tasks/results/_widget_tasks.py
cisaacstern: @atmorling yes good observation. You are correct that the create map and create plot tasks are nearly identical, with the exception of setting `widget_type="map"` or `widget_type="plot"`. And therefore could be combined into a single `create_html_widget_single_view` as long as `widget_type` (with options "map" or "plot") was also a parameter. If this ultimately seems better, I'm open to considering this. The reason I've kept them separate (at the cost of offending DRY sensibilities 😅 ) is to keep the compilation spec a little more descriptive, i.e.

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/4f63f8cef8b61f2cf9e473ce426b3aba0534d449/examples/compilation-specs/time-density.yaml#L11-L16

This ensures that the widget resulting from `create_map_widget_single_view` will have the correct value of `widget_type="map"` (since it's hardcoded into that task).

If we instead had 
```yaml
draw_ecomap: 
  geodataframe: calculate_time_density 
persist_text: 
  text: draw_ecomap 
create_html_widget_single_view: 
  data: persist_text 
```

Then that would require the user to know that they would also have to set `widget_type="map"` as a parameter. Currently, that happens in a totally different place (the web form) which seemed very odd). But having gotten to this point in my explanation (and I truly didn't anticipate getting here, so thanks for the question!) ... I'm now realizing that once #90 is finished, we have a path to setting explicit/literal values (that is, not just results from other tasks) in the yaml spec directly, so we could then do something like:

```yaml
- name: Persist HTML Text for Map
  id: save_map_html
  task: persist_text 
  with:
    text: ${{ ecomap.return }}
- name: Create Map Widget
  id: create_map_widget
  task: create_html_widget_single_view 
  with:
    widget_type: "map"
    data: ${{ save_map_html.return }}
```
Most of this ☝️ (`name`/`id`/`task`/`with` field names, etc.) is already encoded in #90. The new thing here is the idea that a literal value (`"map"`) could also be set in the spect itself. I've been wondering if we would encounter situations where we'd want to do this, but just hadn't run into one yet. This seems like a good motivator for that. What do you think? I'll open an issue to track this.

cisaacstern: Opened #102 to track

Issue #95, opened 2024-07-23
Remove AST parsing in Jupytext target in favor of calling tasks directly


cisaacstern:
In #91 I am finding that writing the widget tasks in such a way that they are compatible with the AST parsing logic required by the current implementation of the Jupytext compilation target, is becoming quite a headache.

The reason I used AST at all was to make the Jupytext notebook cells reflect the code in the _body_ of the task (rather than calling the task function directly). At the time I first drafted this, this seemed like it would make the notebook target look a little more familiar to users of ecoscope core.

Perhaps that is true, however, I see now that there are a lot of edge cases that are now coming up where this logic reveals itself to be pretty brittle. Specifically in #91, I'm hitting the fact that _every_ object used in the body of the task must either be defined or imported in the body of the task. This worked okay up to this point, but for the case of the widgets PR, it makes sense to define some shared dataclasses at the module top level. This totally standard programming style which is otherwise compatible with everything else we are trying to achieve in workflows is, unfortunately, incompatible with our current (quite simplistic) AST parsing logic.

(I noted in a recent review that @Yun-Wu also encountered a need to shoehorn her code into a format that the AST parsing logic understood, again solely for the purpose of Jupytext compilation.)

If a key near term goal of this package was to generate Jupyter notebooks in the current style, I'm sure we could find a way to maintain the current style. That is not the case, however. Instead, in relation to our primary development goals, I understand the Jupytext outputs to be a nice to have, but not a need to have. And certainly we are not targeting any specific style for these outputs as an explicit product requirement.

With this in mind, I propose that I should unburden our core development goals from this AST logic and rework the Jupytext outputs to just call the tasks directly. This seems to me to be the clearest way to keep the Jupyter outputs without hindering task development unnecessarily. I can draw up a PR tomorrow for review.

cc @walljcg

PR #97, opened 2024-07-23
Remove airflow (and dask) templates and associated infra

### LINKED ISSUE #96: Remove unused templates (airflow and dask) and associated test infra
Of the two templates referenced in the title, Airflow is by far the more developed.

In the interest of keeping this package focused, we should remove these unused components.

If for some reason we decide we want to use any of this code in the future, we can always come back and revive the removed code.

cisaacstern:
Closes #96


cisaacstern:
cc @Yun-Wu and @walljcg for visibility

I am prioritizing this now because:

- I'm working on #91 
- Which is semi-blocked by #95
- In starting to resolve #95, which is more-or-less about simplifying and focusing our compilation targets to better align with our current development goals, it struck me that these unused parts of the package were quickly going stale, and not part of our current focus. I am removing them to make it easier for me (and hopefully others) to filter out the "noise" in this repo and focus on our main development priorities.

cisaacstern:
Merging now to unblock myself. I'm trusting this is uncontroversial because we are not currently using any of this code, but we can always revive any of this code if I've misunderstood something.

walljcg:
Thanks Charles! I fully support dropping any clutter and can re-integrate later if needed.

PR #98, opened 2024-07-23
Drop AST parsing for Jupytext target in favor of calling task functions directly

### LINKED ISSUE #95: Remove AST parsing in Jupytext target in favor of calling tasks directly

cisaacstern:
Closes #95

As detailed in #95, this unblocks #91 by allowing us to avoid some gnarly workarounds that were becoming necessary there _solely_ for the purpose of accommodating AST parsing as it was implemented for Jupytext compilation.

Again, as discussed in #95, if a central goal of this package was to provide AST parsing, then we could iterate on that feature some more. As it stands, however, that was more of a (frankly) unconsidered experimental idea, that is not tied to our core objectives, and is now becoming a major burden to support.

Importantly, please note that this PR _**does not**_ remove the Jupytext (and thereby Jupyter) compilation target. We still have that! Instead, it just (drastically) simplifies the nature of the code put into that target, bringing it into closer alignment with the code used in the sequential script.

Arguably, this is actually a big win for maintainability, since it reduced divergence in our compilation targets, making them much more similar to one-another, which I actually find that I like aesthetically and from a clarity standpoint (apart from the fact that it is also dramatically easier to maintain).

### DIFF
--- a/ecoscope_workflows/registry.py
+++ b/ecoscope_workflows/registry.py
@@ -3,13 +3,12 @@ entry points. This design is heavily inspired by the `fsspec` package's `registr
 to which we owe a debt of gratitude.
 """
 
-import ast
 import types
 from dataclasses import dataclass
 from enum import Enum
 from importlib import import_module
 from importlib.metadata import entry_points
-from inspect import getmembers, getsource, ismodule
+from inspect import getmembers, ismodule
 from typing import Annotated, Any, Generator, get_args
 
 import ruamel.yaml
@@ -19,7 +18,6 @@ from pydantic import (
     Field,
     FieldSerializationInfo,
     TypeAdapter,
-    computed_field,
     field_serializer,
 )
 from pydantic.functional_validators import AfterValidator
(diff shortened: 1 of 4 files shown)


Yun-Wu (approved):
LGTM

walljcg:
This is great Charles! Thank-you!

PR #104, opened 2024-07-26
YAML spec vscode extension for realtime feedback

### LINKED ISSUE #103: VSCode Extension for live validation of compilation specs
#90 makes the YAML spec more powerful but also more complex.

Pydantic will validate the specs on parse time, but I am finding that having a VS Code extension that does that dynamically as I type probably would be a big productivity boost.

And if I need that (as the person designing this) then probably others might benefit from it too.

cisaacstern:
Closes #103 

🤯 

Screenshot 2024-07-26 at 3 46 20 PM


cisaacstern:
Some next steps:
- [ ] Get language server to start with the currently selected python interpreter in vscode
- [ ] Setup a "hot reload" configuration of the server so that the validation (optionally) evolves as the the Spec model evolves
- [ ] Document an easy way to set this up in your local environment (eventually could be released on marketplace, but that's probably overkill at this point)
- [ ] Use the [`loc` field in the error](https://docs.pydantic.dev/latest/errors/errors/) to backtrack to which _**line(s)**_ (and possibly columns as well) are causing the error (rather than red-underlining the whole text)
- [ ] Figure out how to have this validation only ecoscope compilation specs (not a _all_ yaml files)

PR #120, opened 2024-07-31
Set groupers, split groups (and mode="mapvalues")

### LINKED ISSUE #72: Task: Define Groupers
Define groupers

https://lucid.app/lucidchart/648e2fe6-0416-4688-912c-d44f14297372/edit?invitationId=inv_83b12d9a-57da-4ce0-b28f-c5957b4303b3&referringApp=slack&page=0_0#

### LINKED ISSUE #109: Task: groupby (i.e. "split")
This task uses pandas.Dataframe.groupby internally to apply the groupers defined by #72, and split an input dataframe into groups for processing.

It may be easiest to do this together with #72 since these features are closely linked.

Depending on the setting (local, cloud) and/or workflow configuration, this task may emit either actual dataframe groups, or possibly references (urls) to serialized groups (i.e. hive-partitioned parquet). The latter is demonstrated in #45.

We should also leave room for the eventual scenario in which the data may be live in a backed that supports SQL queries (BigQuery, DuckDB, etc.), in which case the groupby may not need to happen in-memory. This is not relevant to our immediate concerns but, if there's a way to do so that doesn't take much extra time, it may be useful to make a nod to this idea somewhere in this implementation (a NotImplemented code path, etc.) so we keep it in mind.

cisaacstern:
WIP

Closes #72 
Closes #109 

As noted in the first commit, I believe to achieve the style prototyped in #45, this also means adding a `mode="mapvalues"`. I will explain my thinking on this more if/when I convince myself that it is indeed necessary.

### DIFF
--- a/.github/workflows/unit-test.yaml
+++ b/.github/workflows/unit-test.yaml
@@ -27,6 +27,10 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -e ".[test]"
 
+      - name: Run doctests
+        run: |
+          python3 -m pytest -v ecoscope_workflows/ --doctest-modules
+
       - name: Test with pytest
         env:
           EARTHRANGER_SERVER: ${{ secrets.EARTHRANGER_SERVER }}
--- a/ecoscope_workflows/annotations.py
+++ b/ecoscope_workflows/annotations.py
@@ -2,6 +2,7 @@ from inspect import ismethod
 from typing import Annotated, Any, TypeVar, get_args, get_origin
 
 import pandera as pa
+import pandera.typing as pa_typing
 from pydantic import GetJsonSchemaHandler
 from pydantic.functional_validators import BeforeValidator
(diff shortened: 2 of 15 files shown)


cisaacstern:
some notes to self on next steps:

- [ ] use the ArgName given in `iter:` to assign iterator variable name in `map` and `mapvalues`, this is more readable (as in last commit for `mapvalues`)
- [ ] use [N] indexing to indicate which parts of the per-element return should be passed where (again, see last commit)
- [ ] with these two pieces implemented in template + compiler, we should be able to recreate the current mapvalues example script from the the current mapvalues example spec

this stuff is hard to visualize in one's head ... it will work, but a reviving #26 would probably be helpful for communication

cisaacstern:
note to self: maybe use a classmethod on the variable type for structural pattern matching of square bracket indexing 🤔

cisaacstern:
superseded by #160

Issue #93, opened 2024-07-22
Review task params against standard REACT JSON Schema forms


walljcg:
Review all the task params within the patrols workflow and compare against the 'out-of-the-box' (shot-cn) json schema controls to see what we might need to customize ourselves versus what we get for free (e.g.,  floats, text, arrays, tuples, dicts, datetime, integers, )

cisaacstern:
xref https://github.com/wildlife-dynamics/ecoscope-workflows/issues/82#issuecomment-2240880664

Yun-Wu:
Here are the issues identified during research:

- `dict` is not supported by react-jsonschema-form. While we could represent it as an array of key-value pairs, our parser has to understand the key-value types, while dictionary values can be `Any`. `style_kws` is a good example where lonboard supports arguments of any types but we need to parse user input first to extract values of the correct type. My recommendation is to only support a subset of arguments which we specify the types, like what we did [here](https://github.com/wildlife-dynamics/ecoscope-modules/blob/main/ecoscope_modules/configs/module_configs/map_config.py). Happy to hear your thoughts: @walljcg @cisaacstern 
- `tuple` is not supported either, (used by `Filter`). Any reason we didn't make it a pydantic data model?
- `get_params_jsonschema` returns a dict of json schemas while react-jsonschema-form requires fields like `properties`. Added those fields in the dictionary.
- Added default values for some tasks.
- Task ID is not user-friendly in the forms. Use task names instead.
- More restrictions need to be applied in `patrol_workflow`
  - Filter point coordinates: like this: [[180, 90], [0, 0]]
- Other things that could be polished
  - Hide items with defaults
  - Nullable fields like `filename`, `view`, `groupers`
 
The fix can be found here #123

cisaacstern:
Responses to above points in https://github.com/wildlife-dynamics/ecoscope-workflows/pull/123#pullrequestreview-2216061317 (responded on the PR to combine that with my review there)

Yun-Wu:
Created the following tasks based on our discussion

- #130 
- #131 
- #132 
- #133 
- #134 

Hopefully this week I can get all patrol-workflow configuration correctly rendered (stretch goal). Then I'll work on key improvement for user experience.

PR #123, opened 2024-08-02
Generate react-jsonschema-form style

Yun-Wu:
- **Generate react-jsonschema-form**
- **default value**
- **enum**
- **fix unittest**

### DIFF
--- a/ecoscope_workflows/compiler.py
+++ b/ecoscope_workflows/compiler.py
@@ -17,6 +17,7 @@ from pydantic import (
 )
 from pydantic.functional_validators import AfterValidator, BeforeValidator
 
+from ecoscope_workflows.jsonschema import ReactJSONSchemaFormConfiguration
 from ecoscope_workflows.registry import KnownTask, known_tasks
 
 TEMPLATES = pathlib.Path(__file__).parent / "templates"
@@ -394,13 +395,19 @@ class DagCompiler(BaseModel):
         )
 
     def get_params_jsonschema(self) -> dict[str, Any]:
-        schema: dict[str, Any] = {}
-        schema["title"] = "Ecoscope Workflow Configurations Form"
-        schema["properties"] = {
-            t.name: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
-            for t in self.spec.workflow
-        }
-        return schema
+        return ReactJSONSchemaFormConfiguration(
+            properties={
+                t.name: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
+                for t in self.spec.workflow
+            }
+        ).model_dump()
+        # schema: dict[str, Any] = {}
+        # schema["title"] = "Ecoscope Workflow Configurations Form"
+        # schema["properties"] = {
+        #     t.name: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
+        #     for t in self.spec.workflow
+        # }
+        # return schema
 
     def get_params_fillable_yaml(self) -> str:
         yaml_str = ""
--- a/ecoscope_workflows/compiler.py
+++ b/ecoscope_workflows/compiler.py
@@ -397,7 +397,7 @@ class DagCompiler(BaseModel):
         schema: dict[str, Any] = {}
         schema["title"] = "Ecoscope Workflow Configurations Form"
         schema["properties"] = {
-            t.id: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
+            t.name: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
             for t in self.spec.workflow
         }
         return schema
--- a/ecoscope_workflows/compiler.py
+++ b/ecoscope_workflows/compiler.py
@@ -3,14 +3,14 @@ import functools
 import keyword
 import pathlib
 import subprocess
-from typing import Annotated, Callable, Literal, TypeAlias
+from typing import Annotated, Any, Callable, Literal, TypeAlias
 
 from jinja2 import Environment, FileSystemLoader
 from pydantic import (
     BaseModel,
     ConfigDict,
-    PlainSerializer,
     Field,
+    PlainSerializer,
     computed_field,
     model_serializer,
     model_validator,
@@ -18,7 +18,6 @@ from pydantic import (
 from pydantic.functional_validators import AfterValidator, BeforeValidator
 
 from ecoscope_workflows.registry import KnownTask, known_tasks
-
 
 TEMPLATES = pathlib.Path(__file__).parent / "templates"
 
@@ -394,11 +393,14 @@ class DagCompiler(BaseModel):
             + [arg for t in self.spec.workflow for arg in t.map_iterable]
         )
 
-    def get_params_jsonschema(self) -> dict[str, dict]:
-        return {
+    def get_params_jsonschema(self) -> dict[str, Any]:
+        schema: dict[str, Any] = {}
+        schema["title"] = "Ecoscope Workflow Configurations Form"
+        schema["properties"] = {
             t.id: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
             for t in self.spec.workflow
         }
+        return schema
 
     def get_params_fillable_yaml(self) -> str:
         yaml_str = ""
--- a/ecoscope_workflows/compiler.py
+++ b/ecoscope_workflows/compiler.py
@@ -401,13 +401,6 @@ class DagCompiler(BaseModel):
                 for t in self.spec.workflow
             }
         ).model_dump()
-        # schema: dict[str, Any] = {}
-        # schema["title"] = "Ecoscope Workflow Configurations Form"
-        # schema["properties"] = {
-        #     t.name: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
-        #     for t in self.spec.workflow
-        # }
-        # return schema
 
     def get_params_fillable_yaml(self) -> str:
         yaml_str = ""
--- a/ecoscope_workflows/jsonschema.py
+++ b/ecoscope_workflows/jsonschema.py
@@ -88,3 +88,8 @@ class ReactJSONSchemaFormFilters(BaseModel):
     @model_serializer
     def ser_model(self) -> dict[str, Any]:
         return {"schema": self._schema}
+
+
+class ReactJSONSchemaFormConfiguration(BaseModel):
+    title: str = "Ecoscope Workflow Configurations Form"
+    properties: dict[str, Any]
--- a/ecoscope_workflows/tasks/io/_earthranger.py
+++ b/ecoscope_workflows/tasks/io/_earthranger.py
@@ -37,12 +37,12 @@ def get_subjectgroup_observations(
     subject_group_name: Annotated[
         str, Field(description="Name of EarthRanger Subject")
     ],
+    since: Annotated[datetime, Field(description="Start date")],
+    until: Annotated[datetime, Field(description="End date")],
     include_inactive: Annotated[
         bool,
         Field(description="Whether or not to include inactive subjects"),
-    ],
-    since: Annotated[datetime, Field(description="Start date")],
-    until: Annotated[datetime, Field(description="End date")],
+    ] = True,
 ) -> DataFrame[SubjectGroupObservationsGDFSchema]:
     """Get observations for a subject group from EarthRanger."""
     return client.get_subjectgroup_observations(
--- a/ecoscope_workflows/tasks/io/_earthranger.py
+++ b/ecoscope_workflows/tasks/io/_earthranger.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import Annotated, Any
+from typing import Annotated, Any, Literal
 
 import pandas as pd
 import pandera as pa
@@ -64,10 +64,8 @@ def get_patrol_observations(
         Field(description="list of UUID of patrol types"),
     ],
     status: Annotated[
-        list[str],
-        Field(
-            description="list of 'scheduled'/'active'/'overdue'/'done'/'cancelled'",
-        ),
+        list[Literal["active", "overdue", "done", "cancelled"]],
+        Field(description="list of 'scheduled'/'active'/'overdue'/'done'/'cancelled'"),
     ],
     include_patrol_details: Annotated[
         bool, Field(default=False, description="Include patrol details")
--- a/ecoscope_workflows/tasks/preprocessing/_preprocessing.py
+++ b/ecoscope_workflows/tasks/preprocessing/_preprocessing.py
@@ -1,12 +1,12 @@
-from typing import Any, Annotated
+from typing import Annotated, Any
 
 import pandas as pd
 import pandera as pa
 from pydantic import Field
 
+from ecoscope_workflows.annotations import DataFrame, JsonSerializableDataFrameModel
 from ecoscope_workflows.decorators import distributed
 from ecoscope_workflows.tasks.io import SubjectGroupObservationsGDFSchema
-from ecoscope_workflows.annotations import JsonSerializableDataFrameModel, DataFrame
 
 
 class RelocationsGDFSchema(SubjectGroupObservationsGDFSchema):
@@ -20,7 +20,7 @@ def process_relocations(
     filter_point_coords: Annotated[list[list[float]], Field()],
     relocs_columns: Annotated[list[str], Field()],
 ) -> DataFrame[RelocationsGDFSchema]:
-    from ecoscope.base import RelocsCoordinateFilter, Relocations
+    from ecoscope.base import Relocations, RelocsCoordinateFilter
 
     relocs = Relocations(observations)
 
@@ -63,15 +63,14 @@ class TrajectoryGDFSchema(JsonSerializableDataFrameModel):
 @distributed
 def relocations_to_trajectory(
     relocations: DataFrame[RelocationsGDFSchema],
-    min_length_meters: Annotated[float, Field()],
-    max_length_meters: Annotated[float, Field()],
-    max_time_secs: Annotated[float, Field()],
-    min_time_secs: Annotated[float, Field()],
-    max_speed_kmhr: Annotated[float, Field()],
-    min_speed_kmhr: Annotated[float, Field()],
+    min_length_meters: Annotated[float, Field()] = 0.1,
+    max_length_meters: Annotated[float, Field()] = 10000,
+    max_time_secs: Annotated[float, Field()] = 3600,
+    min_time_secs: Annotated[float, Field()] = 1,
+    max_speed_kmhr: Annotated[float, Field()] = 120,
+    min_speed_kmhr: Annotated[float, Field()] = 0.0,
 ) -> DataFrame[TrajectoryGDFSchema]:
-    from ecoscope.base import Relocations
-    from ecoscope.base import Trajectory, TrajSegFilter
+    from ecoscope.base import Relocations, Trajectory, TrajSegFilter
 
     # trajectory creation
     traj = Trajectory.from_relocations(Relocations(relocations))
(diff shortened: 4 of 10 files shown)


cisaacstern (commented):
This is amazing 🙏 

Responses below to your comment https://github.com/wildlife-dynamics/ecoscope-workflows/issues/93#issuecomment-2264480217

> - `dict` is not supported by react-jsonschema-form. While we could represent it as an array of key-value pairs, our parser has to understand the key-value types, while dictionary values can be `Any`. `style_kws` is a good example where lonboard supports arguments of any types but we need to parse user input first to extract values of the correct type. My recommendation is to only support a subset of arguments which we specify the types, like what we did [here](https://github.com/wildlife-dynamics/ecoscope-modules/blob/main/ecoscope_modules/configs/module_configs/map_config.py). Happy to hear your thoughts: @walljcg @cisaacstern 


As we discussed in our sync, if the `dict` is supposed to be user-configurable (as is `style_kws`), then yes I agree that a pydantic model makes perfect sense. Here's the scratch code from our sync that shows unpacking using `.model_dump()` so that the model is just for parsing inputs and the ecoscope core code only ever "sees" a dict:

💡 _**Edit**_: See comment https://github.com/wildlife-dynamics/ecoscope-workflows/pull/123#issuecomment-2265901807 below for a correction to the code below

```python
class StyleKws(BaseModel):
    a: int
    b: str


@distributed
def draw_ecomap(
    ...
    style_kws: Annotated[StyleKws ...,
    ...
) -> Annotated[str, Field()]:
    ...
            # 💡 Edit: See comment below for a correction to this
            # https://github.com/wildlife-dynamics/ecoscope-workflows/pull/123#issuecomment-2265901807
            m.add_scatterplot_layer(geodataframe, **style_kws.model_dump())
```

> - `tuple` is not supported either, (used by `Filter`). Any reason we didn't make it a pydantic data model?

As we discussed in sync, there are certain fields, including `Filter`, which I've mostly convinced myself we never want to expose to the user, because they are used to manage DAG operations (like groupby/apply keys, etc.) ... for these cases maybe we can have a custom `Field` constructor like this: 

```python
# ecoscope_workflows/annotations.py

from pydantic import Field 

def UserFormExcludeField(*args, **kwargs):
    return Field(*args, **kwargs, exclude=True)
```
Of course `exclude=True` can be set using pydantic's built-in Field, but I like the idea of this being more readable on the page. (We called this a `SpecOnlyField` in our sync but maybe the name above is more descriptive?)


> - `get_params_jsonschema` returns a dict of json schemas while react-jsonschema-form requires fields like `properties`. Added those fields in the dictionary.

In the sync you asked for me to take a look at this implementation and were wondering if it would be better to use a model rather than just modifying the `schema` dict ...

Good news! We already have this model which I believe can be easily generalized to serve this use case:

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/4ed2ec81f785fcd1d00294fb47679f131dadfd71/ecoscope_workflows/jsonschema.py#L73-L77

Please feel free to change that as needed. It is used by the dashboard tasks to create the JSON model for the filters. So changes here will require some small changes in the dashboard code and tests. This would make the current diff a lot bigger (though not necessarily a lot more complicated) ... so I would also 💯 endorse the idea of opening a new ticket and doing this as a separate PR.

> - Added default values for some tasks.

👍 

> - Task ID is not user-friendly in the forms. Use task names instead.

Agree 👍 

> - More restrictions need to be applied in `patrol_workflow`
>   - Filter point coordinates: like this: [[180, 90], [0, 0]]

Are there remaining design questions as to how to do this?

> - Other things that could be polished
>   - Hide items with defaults

As we dicussed in sync, I'm not sure the best way to handle this in the JSON yet, but for now I would suggest we add an additional custom Field constructor as a placeholder for how we will implement that logic, such as:

```python
# ecoscope_workflows/annotations.py

from pydantic import Field

def AdvancedField(*args, **kwargs):
    return Field(
        *args,
        **kwargs,
        # TODO: figure out the best way to communicate "advanced" fields to the UI
        # via the JSON schema. {"advanced": True} is a placeholder for now.
        # json_schema_extra={"advanced": True},
    )
```

Once we resolve the `TODO` comment there, all fields annotated as

```python
def f(
   arg: Annotated[..., AdvancedField(...),
): ...
```

will have that change applied 😃 


>   - Nullable fields like `filename`, `view`, `groupers`
 
Not all nullable fields are `UserFormExcludeFields`, but as I am thinking about it, I believe `UserFormExcludeField` can be used for all of the nullable fields you mention here.

cisaacstern:
> As we discussed in our sync, if the dict is supposed to be user-configurable (as is style_kws), then yes I agree that a pydantic model makes perfect sense. Here's the scratch code from our sync that shows unpacking using .model_dump() so that the model is just for parsing inputs and the ecoscope core code only ever "sees" a dict:


Oh! I just realized a problem with the code example above, which is that outside the `validate_call`/`.replace(validate=True)` calling context (for example, in our unit tests, or in the jupytext notebooks we've been generating), we cannot assume that the value of an argument annotated with a BaseModel will in fact _be_ a BaseModel in the body of the function. That is because without `validate_call`, pydantic does not parse inputs.

There is a way around this, though, which I think makes the code in the body of the ask cleaner. First, to establish a baseline, this illustrates the problem I am referring to:

```python
from typing import Annotated
from pydantic import BaseModel, validate_call, Field

class StyleKws(BaseModel):
    a: int
    b: str

def f(s: Annotated[StyleKws, Field(description="styling")]):
    return s.model_dump()


style_kws_form_data = {"a": 1, "b": "2"}

print(validate_call(f)(s=style_kws_form_data))
# >> {'a': 1, 'b': '2'}
f(s=style_kws_form_data)
# >> raises `AttributeError: 'dict' object has no attribute 'model_dump'`
```

To solve this we can define an "after validator" which, in the Pydantic `validate_call` context, will dump the model to a dict for us before passing it to the body of the task:

```python
from pydantic.functional_validators import AfterValidator

def f(
    s: Annotated[
        StyleKws,
        Field(description="styling"),
        # as the name suggestions, this runs _after_ the input dict is validated
        AfterValidator(lambda m: m.model_dump())
     ]
):
    return s

# this still works
print(validate_call(f)(s=style_kws_form_data))
# >> {'a': 1, 'b': '2'}

# but now this also works


# and we still get validation with pydantic
invalid_form_data = {"a": [1, 2, 3], "b": "2"}
validate_call(f)(s=invalid_form_data)
# >> raises `ValidationError: 1 validation error for f s.a
#  Input should be a valid integer [type=int_type, input_value=[1, 2, 3], input_type=list]

# or not, without pydantic
f(s=invalid_form_data)
# >> {'a': [1, 2, 3], 'b': '2'}
```

cisaacstern:
Lint workflow issue is fixed by #129.

P.S., sorry for the merge conflicts from #117. They should hopefully not be too bad to resolve by:
- ❌ Deleting `examples/dags/time_density_dag.jupytext.py`
- ❌ Deleting `examples/params/time_density_params_fillable.json`
- ♻️ Recompiling and fixing conflicts in  `examples/params/patrol_workflow_params_fillable.json
`

Yun-Wu:
Thanks for your detailed response @cisaacstern . Based on our discussion, I've created the following tasks:

- #130 
- #131 
- #132 
- #133 
- #134 

The challenge in reusing `ReactJSONSchemaFormFilters` is the absence of `RJSFFilter`. I considered the same approach initially, but currently, the JSON schema for each task is generated by `TypeAdapter` of a `func`. Thus, we have to create a data model for each task, which will break auto-generation of configuration from a function signature. Any thoughts?

Yun-Wu:
> > As we discussed in our sync, if the dict is supposed to be user-configurable (as is style_kws), then yes I agree that a pydantic model makes perfect sense. Here's the scratch code from our sync that shows unpacking using .model_dump() so that the model is just for parsing inputs and the ecoscope core code only ever "sees" a dict:
> 
> Oh! I just realized a problem with the code example above, which is that outside the `validate_call`/`.replace(validate=True)` calling context (for example, in our unit tests, or in the jupytext notebooks we've been generating), we cannot assume that the value of an argument annotated with a BaseModel will in fact _be_ a BaseModel in the body of the function. That is because without `validate_call`, pydantic does not parse inputs.
> 
> There is a way around this, though, which I think makes the code in the body of the task cleaner. First, to establish a baseline, this illustrates the problem I am referring to:
> 
> ```python
> from typing import Annotated
> from pydantic import BaseModel, validate_call, Field
> 
> class StyleKws(BaseModel):
>     a: int
>     b: str
> 
> def f(s: Annotated[StyleKws, Field(description="styling")]):
>     return s.model_dump()
> 
> 
> style_kws_form_data = {"a": 1, "b": "2"}
> 
> print(validate_call(f)(s=style_kws_form_data))
> # >> {'a': 1, 'b': '2'}
> f(s=style_kws_form_data)
> # >> raises `AttributeError: 'dict' object has no attribute 'model_dump'`
> ```
> 
> To solve this we can define an "after validator" which, in the Pydantic `validate_call` context, will dump the model to a dict for us before passing it to the body of the task:
> 
> ```python
> from pydantic.functional_validators import AfterValidator
> 
> def f(
>     s: Annotated[
>         StyleKws,
>         Field(description="styling"),
>         # as the name suggests, this runs _after_ the input dict is validated
>         AfterValidator(lambda m: m.model_dump())
>      ]
> ):
>     return s
> 
> # this still works
> print(validate_call(f)(s=style_kws_form_data))
> # >> {'a': 1, 'b': '2'}
> 
> # but now this also works (whereas before we got an AttributeError)
> print(f(s=style_kws_form_data))
> # >> {'a': 1, 'b': '2'}
> 
> # and we still get validation with pydantic
> invalid_form_data = {"a": [1, 2, 3], "b": "2"}
> validate_call(f)(s=invalid_form_data)
> # >> raises `ValidationError: 1 validation error for f s.a
> #  Input should be a valid integer [type=int_type, input_value=[1, 2, 3], input_type=list]
> 
> # or not, without pydantic
> f(s=invalid_form_data)
> # >> {'a': [1, 2, 3], 'b': '2'}
> ```

To solve this issue, I propose passing in parameters with the pydantic data model type directly like this:
```
from typing import Annotated
from pydantic import BaseModel, validate_call, Field

class StyleKws(BaseModel):
    a: int
    b: str

def f(s: Annotated[StyleKws, Field(description="styling")]):
    return s.model_dump()


style_kws_from_serialized_data = {"a": 1, "b": "2"}
style_kws_direct_function_call = StyleKws(a=1, b="2")

# This should work
print(validate_call(f)(s=style_kws_form_data))

# This should also work
f(s=style_kws_form_data)
```

Otherwise it confuses users who are directly calling this function since the function signature requires a `StyleKws` type while a `dict` is actually needed.

cisaacstern:
> To solve this issue, I propose passing in parameters with the pydantic data model type directly like this:
> Otherwise it confuses users who are directly calling this function since the function signature requires a StyleKws type while a dict is actually needed.

I agree that my earlier proposal was not ideal for this reason.

I propose the following as a middle ground between my first proposal and your comment above:

https://github.com/wildlife-dynamics/ecoscope-workflows/issues/130#issuecomment-2269286105

With an example of how to use that in a real signature here:
https://github.com/wildlife-dynamics/ecoscope-workflows/pull/135/files#r1704260220

ecoscope_workflows/compiler.py
```diff
@@ -394,11 +393,14 @@ def _omit_args(self) -> list[str]:
...
-        return {
-            t.id: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
+    def get_params_jsonschema(self) -> dict[str, Any]:
+        schema: dict[str, Any] = {}
+        schema["title"] = "Ecoscope Workflow Configurations Form"
+        schema["properties"] = {
+            t.name: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
             for t in self.spec.workflow
         }
+        return schema
```
cisaacstern, line 403:
```suggestion
    def get_params_jsonschema(self) -> dict[str, Any]:
        return ReachJSONSchemaForm(
            title="Ecoscope Workflow Configurations Form"
            properties={
                t.name: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
                for t in self.spec.workflow
            },
        ).model_dump()
```

> The challenge in reusing ReactJSONSchemaFormFilters is the absence of RJSFFilter. I considered the same approach initially, but currently, the JSON schema for each task is generated by TypeAdapter of a func. Thus, we have to create a data model for each task, which will break auto-generation of configuration from a function signature. Any thoughts?


Something like this diff ☝️ is what I was imagining. It would require some generalization of the RJSF* models...

cisaacstern: These models 

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/4ed2ec81f785fcd1d00294fb47679f131dadfd71/ecoscope_workflows/jsonschema.py#L66-L74

could be generalized something like ...

```diff
-class RJSFFilter(BaseModel): 
-   """Model representing a React JSON Schema Form filter.""" 
+class RJSFField(BaseModel): 
+   """Model representing a React JSON Schema Form.""" 

-   property: RJSFFilterProperty
-   uiSchema: RJSFFilterUiSchema
+   property: RJSFFilterProperty | dict
+   uiSchema: RJSFFilterUiSchema | dict


-class ReactJSONSchemaFormFilters(BaseModel): 
-    options: dict[str, RJSFFilter] 
+class ReactJSONSchemaFormFilters(BaseModel):
+    title: str
+    properties: dict[str, RJSFField] 

```

This is just some quick pseudocode, happy to sync more on the details if that would be useful.

cisaacstern (commented):
> The challenge in reusing ReactJSONSchemaFormFilters is the absence of RJSFFilter. I considered the same approach initially, but currently, the JSON schema for each task is generated by TypeAdapter of a func. Thus, we have to create a data model for each task, which will break auto-generation of configuration from a function signature. Any thoughts?


The below is roughly what I was thinking. TBH I don't think it's necessary for this PR, but if it seems interesting/beneficial to you, then happy to explore it further.

cisaacstern:
> > To solve this issue, I propose passing in parameters with the pydantic data model type directly like this:
> > Otherwise it confuses users who are directly calling this function since the function signature requires a StyleKws type while a dict is actually needed.
> 
> I agree that my earlier proposal was not ideal for this reason.
> 
> I propose the following as a middle ground between my first proposal and your comment above:
> 
> [#130 (comment)](https://github.com/wildlife-dynamics/ecoscope-workflows/issues/130#issuecomment-2269286105)
> 
> With an example of how to use that in a real signature here: https://github.com/wildlife-dynamics/ecoscope-workflows/pull/135/files#r1704260220

Just closing the loop on this here. As I've already commented in the other linked locations, I take @Yun-Wu's point about simplicity of using the model directly as the type annotation, and not having a conflict between the type annotation and a (separate) `FrontendSchema`. The only "cost" of this (which is not a huge cost, really), is that for these types, in contexts where we:
- might otherwise pass parameters in the form of dicts (unit tests, jupyter)
- and _also_ are not calling the tasks with `.replace(validate=True)` (which invokes pydantic type parsing)

then we need to import and instantiate the models directly as the values for this argument.

One question this raises is about the import patters for those models in this context, since we would have to something like:

```python
from ecoscope_workflows.tasks.results import (
    draw_ecomap,
    EcoMapStyleKws,
)


map = draw_ecomap(..., style_kws=EcoMapStyleKws(...))
```

this is likely to raise an error here

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/e596df3d66159e10f0f059a69b422c957cdf3a7b/ecoscope_workflows/registry.py#L55-L67

☝️ This is where we recursively collect user-facing objects from the task modules, and this will raise an error if objects not of type `DistributedTask` or `JsonSerializableDataFrameModel` are found.

The quick fix here (which I am comfortable with) is
1. Define a common base from which all dict parameter models will inherit, probably `ecoscope_workflows.annotations` is the most logical place for this, something like

  ```python
  # ecoscope_workflows/annotations.py
  
  class DictParameter(BaseModel):
      pass
  ```
2. Add that type to the allowable non-`DistributedTask` types in the task collection here: 
  ```diff
  -   elif issubclass(obj, JsonSerializableDataFrameModel): 
  +   elif issubclass(obj, (JsonSerializableDataFrameModel, DictParameter)): 
           continue
  ```
3. Then in `ecoscope_workflows/tasks/<some_directory>/__init__.py` objects which inherit from `DictParameter` can be exposed as user-facing objects by adding them to `__all__`, making them importable from the user-facing import pattern (e.g., from `ecoscope_workflows.tasks.results` instead of `ecoscope_workflows.tasks.results._ecomap`)

Yun-Wu:
@cisaacstern I didn't run into this problem.

> > > To solve this issue, I propose passing in parameters with the pydantic data model type directly like this:
> > > Otherwise it confuses users who are directly calling this function since the function signature requires a StyleKws type while a dict is actually needed.
> > 
> > 
> > I agree that my earlier proposal was not ideal for this reason.
> > I propose the following as a middle ground between my first proposal and your comment above:
> > [#130 (comment)](https://github.com/wildlife-dynamics/ecoscope-workflows/issues/130#issuecomment-2269286105)
> > With an example of how to use that in a real signature here: https://github.com/wildlife-dynamics/ecoscope-workflows/pull/135/files#r1704260220
> 
> Just closing the loop on this here. As I've already commented in the other linked locations, I take @Yun-Wu's point about simplicity of using the model directly as the type annotation, and not having a conflict between the type annotation and a (separate) `FrontendSchema`. The only "cost" of this (which is not a huge cost, really), is that for these types, in contexts where we:
> 
> * might otherwise pass parameters in the form of dicts (unit tests, jupyter)
> * and _also_ are not calling the tasks with `.replace(validate=True)` (which invokes pydantic type parsing)
> 
> then we need to import and instantiate the models directly as the values for this argument.
> 
> One question this raises is about the import patters for those models in this context, since we would have to something like:
> 
> ```python
> from ecoscope_workflows.tasks.results import (
>     draw_ecomap,
>     EcoMapStyleKws,
> )
> 
> 
> map = draw_ecomap(..., style_kws=EcoMapStyleKws(...))
> ```
> 
> this is likely to raise an error here
> 
> https://github.com/wildlife-dynamics/ecoscope-workflows/blob/e596df3d66159e10f0f059a69b422c957cdf3a7b/ecoscope_workflows/registry.py#L55-L67
> 
> ☝️ This is where we recursively collect user-facing objects from the task modules, and this will raise an error if objects not of type `DistributedTask` or `JsonSerializableDataFrameModel` are found.
> 
> The quick fix here (which I am comfortable with) is
> 
> 1. Define a common base from which all dict parameter models will inherit, probably `ecoscope_workflows.annotations` is the most logical place for this, something like
>    ```python
>    # ecoscope_workflows/annotations.py
>    
>    class DictParameter(BaseModel):
>        pass
>    ```
> 2. Add that type to the allowable non-`DistributedTask` types in the task collection here:
>    ```diff
>    -   elif issubclass(obj, JsonSerializableDataFrameModel): 
>    +   elif issubclass(obj, (JsonSerializableDataFrameModel, DictParameter)): 
>             continue
>    ```
> 3. Then in `ecoscope_workflows/tasks/<some_directory>/__init__.py` objects which inherit from `DictParameter` can be exposed as user-facing objects by adding them to `__all__`, making them importable from the user-facing import pattern (e.g., from `ecoscope_workflows.tasks.results` instead of `ecoscope_workflows.tasks.results._ecomap`)

ecoscope_workflows/compiler.py
Yun-Wu: Thanks. I like this!

ecoscope_workflows/compiler.py
```diff
@@ -394,11 +394,20 @@ def _omit_args(self) -> list[str]:
...
+                for t in self.spec.workflow
+            }
+        ).model_dump()
+        # schema: dict[str, Any] = {}
+        # schema["title"] = "Ecoscope Workflow Configurations Form"
+        # schema["properties"] = {
+        #     t.name: t.known_task.parameters_jsonschema(omit_args=self._omit_args)
+        #     for t in self.spec.workflow
+        # }
+        # return schema
```
cisaacstern, line 410:
```suggestion
```

cisaacstern (approved):
Looks awesome, thanks!

The `ReactJSONSchemaFormConfiguration` model definitely adds readability, nice.

Just one comment about removing a stray comment.

PR #135, opened 2024-08-05
Replace kws with data models

Yun-Wu:
- **Create data models for map properties**
- **test with mode-map**
- **pull  out of the json schema of each task**

### DIFF
--- a/ecoscope_workflows/tasks/results/_ecomap.py
+++ b/ecoscope_workflows/tasks/results/_ecomap.py
@@ -4,11 +4,9 @@ from pydantic import Field
 
 from ecoscope_workflows.annotations import AnyGeoDataFrame
 from ecoscope_workflows.decorators import distributed
-from ecoscope_workflows.results._map_config import (
+from ecoscope_workflows.tasks.results._map_config import (
     LayerStyleProperty,
     NorthArrowProperty,
-    ScaleBarPropertyProperty,
-    TitleProperty,
 )
 
 
@@ -21,10 +19,11 @@ def draw_ecomap(
         Literal["Scatterplot", "Path", "Polygon"],
         Field(description="The type of visualization."),
     ],
-    style_kws: Annotated[
+    # TODO: react-jsonschema-form: select different layer props based on the input of data_type
+    style_props: Annotated[
         LayerStyleProperty,
         Field(description="Style arguments for the data visualization."),
-    ],
+    ] = LayerStyleProperty(),
     tile_layer: Annotated[
         str, Field(description="A named tile layer, ie OpenStreetMap.")
     ] = "",
@@ -32,18 +31,10 @@ def draw_ecomap(
         bool, Field(description="Set to true to disable map pan/zoom.")
     ] = False,
     title: Annotated[str, Field(description="The map title.")] = "",
-    title_property: Annotated[
-        TitleProperty,
-        Field(description="Additional arguments for configuring the Title."),
-    ] = {},
-    scale_kws: Annotated[
-        ScaleBarPropertyProperty,
-        Field(description="Additional arguments for configuring the Scale Bar."),
-    ] = {},
-    north_arrow_property: Annotated[
+    north_arrow_props: Annotated[
         NorthArrowProperty,
         Field(description="Additional arguments for configuring the North Arrow."),
-    ] = {},
+    ] = NorthArrowProperty(),
 ) -> Annotated[str, Field()]:
     """
     Creates a map based on the provided layer definitions and configuration.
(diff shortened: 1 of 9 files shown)


ecoscope_workflows/tasks/results/_ecomap.py
```diff
@@ -15,39 +19,34 @@ def draw_ecomap(
...
         Field(description="The type of visualization."),
     ],
-    style_kws: Annotated[
-        dict, Field(description="Style arguments for the data visualization.")
-    ],
+    # TODO: react-jsonschema-form: select different layer props based on the input of data_type
+    style_props: Annotated[
+        LayerStyleProperty,
+        Field(description="Style arguments for the data visualization."),
+    ] = LayerStyleProperty(),
```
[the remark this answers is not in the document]
Yun-Wu: I'm actually worried about the mismatch between the param type and the schema. TypeAdapter is very powerful but can also cause a lot of overhead on how data is serialized and deserialized. The code is also simpler since developers don't need to worry about `FrontendSchema` and potentially `AfterValidator` in the future.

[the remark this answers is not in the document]
cisaacstern: Ok I respect that. Let's go with your proposal then, I am comfortable with that.

Then when we need dicts to pass to `ecoscope` core methods we just call `.model_dump()` there in the body of the task.

Thanks for the thoughtfulness here @Yun-Wu.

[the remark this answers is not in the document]
cisaacstern: Note that this implicates registry task collection, see: https://github.com/wildlife-dynamics/ecoscope-workflows/pull/123#issuecomment-2269623179

ecoscope_workflows/tasks/results/_map_config.py
```diff
@@ -0,0 +1,32 @@
+from pydantic import BaseModel
+
+
+class LayerStyleProperty(BaseModel):
+    auto_highlight: bool = False
+    opacity: float = 1
+    pickable: bool = True
+
+
+class PathLayerProperty(LayerStyleProperty):
```
atmorling, line 10:
```suggestion
class PolylineLayerProperty(LayerStyleProperty):
```
this layer has been renamed in the latest `ecoscope`

ecoscope_workflows/tasks/results/_map_config.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class ShapeLayerProperty(LayerStyleProperty):
+    filled: bool = True
+    get_fill_color: list[int] = [0, 0, 0, 255]
+    get_line_color: list[int] = [0, 0, 0, 255]
+    get_line_width: float = 1
+
+
+class ScatterPlotLayerProperty(ShapeLayerProperty):
```
atmorling, line 22:
```suggestion
class PointLayerProperty(ShapeLayerProperty):
```
Renamed in `ecoscope` Same as above

ecoscope_workflows/tasks/results/_map_config.py
```diff
@@ -0,0 +1,32 @@
...
+class PathLayerProperty(LayerStyleProperty):
+    get_color: list[int] = [0, 0, 0, 255]
+    get_width: float = 1
+
+
+class ShapeLayerProperty(LayerStyleProperty):
+    filled: bool = True
+    get_fill_color: list[int] = [0, 0, 0, 255]
+    get_line_color: list[int] = [0, 0, 0, 255]
+    get_line_width: float = 1
```
atmorling, line 19: I'm not sure how much I love having an extra abstraction here that isn't reflected downstream (or the precedence that sets). I think repetition is fine here for the sake of clarity. Happy to be talked down from that hill though.

ecoscope_workflows/tasks/results/_map_config.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class LayerStyleProperty(BaseModel):
+    auto_highlight: bool = False
+    opacity: float = 1
+    pickable: bool = True
+
+
+class PathLayerProperty(LayerStyleProperty):
+    get_color: list[int] = [0, 0, 0, 255]
```
atmorling, line 11: This allows only a single color per map layer. We definitely want to support having per-row colors as well, something like:

```suggestion
    get_color: list[int] | list[list[int]] = [0, 0, 0, 255]
```

ecoscope_workflows/tasks/results/_map_config.py
cisaacstern: Can you expand on what you mean by this @atmorling ?

atmorling: Just that `ShapeLayer` isn't a thing in the `lonboard` API, and that having these fields duplicated in `ScatterplotLayerProperty`/`PointLayerPropery` would perhaps be clearer.

atmorling: Per an offline discussion with @cisaacstern I've come to the realisation `ShapeLayer` here is actually just fine as it won't be surfaced to a user in any way.

PR #160, opened 2024-08-07
Tasks: set groupers + split groups

### LINKED ISSUE #72: Task: Define Groupers

### LINKED ISSUE #109: Task: groupby (i.e. "split")

### LINKED ISSUE #156: Fix broken pandera dataframe import resolution to assist static type checking

### LINKED ISSUE #161: On get-params, duplicate arg names between tasks can result in unintended omissions
see #160

cisaacstern:
Built on #157, which should go in first

closes #72
closes #109
closes #156 
closes #161

Supercedes #120


cisaacstern:
Momentarily blocked by:

- [ ] On `get-params`, duplicate arg names between tasks can result in unintended omissions

I will fix this now

PR #158, opened 2024-08-07
Exclude fields from json schema

Yun-Wu:
- **skip json schema**
- **fix gather_dashboard issue**
- **exclude views**

PR #162, opened 2024-08-08
Merging staging branch (decorator retool + groupby tasks) into `main`

cisaacstern:
Combined diffs from #157 and #160 together in one branch.

I will wait until #158 goes in to limit merge conflict for that PR.


ecoscope_workflows/tasks/results/_ecomap.py
```diff
@@ -101,6 +101,7 @@ def draw_ecomap(
     if tile_layer:
         m.add_layer(EcoMap.get_named_tile_layer(tile_layer))
 
+    geo_layers = [geo_layers] if not isinstance(geo_layers, list) else geo_layers
```
cisaacstern, line 104: @atmorling just about the only place in this entire 60 file diff that required touching code someone else wrote in a consequential way was here... wanted to let you know it somehow seemed necessary to me to allow singleton `LayerDefinition`s (which get turned into len-1 lists here). This came up because in the revised YAML syntax for parallel operations, I don't have a way to express "take this collection of dataframes, and for each one, pass it to `draw_ecomap` as a single-element list" ... so the alternative is to allow `draw_ecomap` to handle that.

Hoping this seems reasonable enough to you, but if not please let me know.

cisaacstern: Though I'm struggling to think of a specific example off the top of my head, this type of interface feels familiar to me from other Python packages, that is the, "pass an iterable of this thing, or if you only have one of them, don't worry, you can just give it to me as-is" call signature.

atmorling: No issues from me. There's similar patterns downstream in `ecomap`

Issue #119, opened 2024-07-31
Patrols Workflow: Add summary stats single value widgets to dashboard


cisaacstern:
Dependencies:

- [x] #113 
- [x] #117 

Once these are done we can do this!

cisaacstern:
I can work on this Friday after #168 goes in.

PR #168, opened 2024-08-08
Grouped ecomap in patrols workflow (plus tasks  `groupbykey` & `add_temporal_index`)

### LINKED ISSUE #163: Task: `groubykey`
In order to this in the grouped data context

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/420e6053abebe0a7c8386bcb7f791c68f4c10b20/examples/compilation-specs/patrol_workflow.yaml#L42-L49

I believe we will need a `groupbykey` task that does, more or less, the same thing as [Beam's `GroupByKey` transform](https://beam.apache.org/documentation/programming-guide/#groupbykey).

I can work on this tomorrow as part of the effort to add groupers to the patrols workflow.

### LINKED ISSUE #169: Task: Add temporal index
To complete #168, we need a way to get Patrol Trajectories and Patrol Events to have a common index. This will allow them to be grouped using the same groupers.

The easiest one to start with is a temporal (e.g. month) index, which we can do if we have a task that does something like this, from ecoscope core:

https://github.com/wildlife-dynamics/ecoscope/blob/6e7dd1ea782c2eef07d7bf794be76e7d4a077042/ecoscope/base/utils.py#L242-L264

(h/t @walljcg for pointing that out in https://github.com/wildlife-dynamics/ecoscope-workflows/pull/51#issuecomment-2190682497.)

I will do this as part of #168

cisaacstern:
closes #163
also incorporates #170 which closed #169

### DIFF
--- a/ecoscope_workflows/tasks/groupby/__init__.py
+++ b/ecoscope_workflows/tasks/groupby/__init__.py
@@ -1,3 +1,3 @@
-from ._groupby import set_groupers, split_groups
+from ._groupby import groupbykey, set_groupers, split_groups
 
-__all__ = ["set_groupers", "split_groups"]
+__all__ = ["groupbykey", "set_groupers", "split_groups"]
--- a/ecoscope_workflows/tasks/groupby/_groupby.py
+++ b/ecoscope_workflows/tasks/groupby/_groupby.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Annotated
+from typing import Any, Annotated
 
 from pydantic import Field
 
@@ -59,8 +59,9 @@ def set_groupers(
     return groupers
 
 
-KeyedIterable = list[tuple[CompositeFilter, AnyDataFrame]]
-CombinedKeyedIterable = list[tuple[CompositeFilter, list[AnyDataFrame]]]
+KeyedIterableOfDataFrames = list[tuple[CompositeFilter, AnyDataFrame]]
+KeyedIterableOfAny = list[tuple[CompositeFilter, Any]]
+CombinedKeyedIterable = list[tuple[CompositeFilter, list[Any]]]
 
 
 @task
@@ -70,7 +71,7 @@ def split_groups(
         list[Grouper], Field(description="Index(es) and/or column(s) to group by")
     ],
 ) -> Annotated[
-    KeyedIterable,
+    KeyedIterableOfDataFrames,
     Field(
         description="""\
         List of 2-tuples of key:value pairs. Each key:value pair consists of a composite
@@ -90,7 +91,7 @@ def split_groups(
 @task
 def groupbykey(
     iterables: Annotated[
-        list[KeyedIterable], Field(description="List of keyed iterables")
+        list[KeyedIterableOfAny], Field(description="List of keyed iterables")
     ],
 ) -> Annotated[
     CombinedKeyedIterable,
--- a/ecoscope_workflows/tasks/groupby/_groupby.py
+++ b/ecoscope_workflows/tasks/groupby/_groupby.py
@@ -59,6 +59,10 @@ def set_groupers(
     return groupers
 
 
+KeyedIterable = list[tuple[CompositeFilter, AnyDataFrame]]
+CombinedKeyedIterable = list[tuple[CompositeFilter, list[AnyDataFrame]]]
+
+
 @task
 def split_groups(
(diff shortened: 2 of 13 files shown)


cisaacstern:
@Yun-Wu @atmorling @walljcg thought I'd pause (at least until after Friday standup) for review, in case you have time to look at this. A few notes/questions:

- We probably want to have #171 for MVP?
- At least using the example data we have here already, it was necessary (for either trajectories or events, can't remember which), to use [`pd.to_datetime` like so](https://github.com/wildlife-dynamics/ecoscope-workflows/pull/168/files#diff-d31de88615af0a451a2545f8bc2e5dd828e2cd31d05c4cedf6af45f18ae59ac4R97-R98). I assume this has some relation to https://github.com/wildlife-dynamics/ecoscope/issues/116? What do you all think about doing this as part of `add_temporal_index` vs. a separate task (either in general, or in terms of whether that's okay for now)? Also, I'm aware `"mixed"` is perhaps an unsafe choice for default arg here, but it was the one that seemed to most readily unblock the remainder of the work here.
- As noted in #163, the affect of `groupbykey` is nearly identical to that of [Beam's GroupByKey transform](https://beam.apache.org/documentation/programming-guide/#groupbykey). The reason we need that here, is that it let's us do this:

  ```
          [trajectories_df]                         [events_df]              
                |                                       |
           split by month                         split by month
            /           \                           /         \   
   ["jan": df0]       ["feb": df1]            ["jan": df2]    ["feb": df3] 
          |                |                        |               |
  create_map_layer    create_map_layer       create_map_layer   create_map_layer
          |                |                        |               |
  ["jan": layer0]     ["feb": layer1]        ["jan": layer2]    ["feb": layer3] 
          |                |                        |               | 
          |                |                        |               |
          └──--------------└──--------------------──┘---------------──┘
                                    |
                               groupbykey
                               /        \
      ["jan": [layer0, layer2]]        ["feb": [layer1, layer3]]
                |                                     |
            draw_ecomap                           draw_ecomap
                |                                     |
  Combined trajectories and                     Combined trajectories and
  events ecomap for january                     events ecomap for february
  ``` 

- After this goes in, I think #119 is the last thing to get us to rough parity with the design in https://github.com/wildlife-dynamics/ecoscope-web/issues/75#issuecomment-2251781687 ?


There's obviously a lot more polishing to be done, but we're getting a lot closer! Thanks in advance for any feedback you may have.

walljcg:
Thanks Charles! This is making a lot of sense! Love the ASCII !!

 groupbykey() is similar to 'collect()' in pyspark? 

As for needing to use pd.to_datetime(), I agree with @atmorling that the 'clean_time_cols()' function should really catch and convert any time cols now since [#231](https://github.com/wildlife-dynamics/ecoscope/pull/231) was merged. There should be 'time', 'created_at', 'updated_at' at the least.

atmorling:
> * The only time column I could find on the patrol events example data we are using here was [`updated_at`, which doesn't seem right](https://github.com/wildlife-dynamics/ecoscope-workflows/pull/168/files#diff-94d7ea8426cfb717be3b639d712b7d1059b826ed2926f2205fd19069ff0b0491R68)? Are there perhaps additional time fields for patrol events we can fetch from ER that are just not in the example file?

@cisaacstern is right here in that currently the only timestamp we have in a `get_patrol_events` response is `update_at`, which is pulled from the `patrol_segments` of an `/acitivity/patrols` response. One option that might be better is to use the timestamp nested in `patrol_segments.geojson.properties`

cisaacstern:
Thanks both for the reviews.

> As for needing to use pd.to_datetime(), I agree with @atmorling that the 'clean_time_cols()' function should really catch and convert any time cols now since https://github.com/wildlife-dynamics/ecoscope/pull/231 was merged.

Good to know for Earthranger data via ecoscope core. Though `add_temporal_index` should be general enough to accommodate non-Earthranger tables (which may need `to_datetime`)... so for now I will leave that in there. Though we may likely want to revisit the defaults.

> There should be 'time', 'created_at', 'updated_at' at the least.

Seems like perhaps not? Now tracking this in #173.

> groupbykey() is similar to 'collect()' in pyspark?

No I think the best pyspark analog is:

https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.RDD.groupByKey.html?highlight=groupbykey#pyspark.RDD.groupByKey

PR #170, opened 2024-08-08
Task: `add_temporal_index`

### LINKED ISSUE #169: Task: Add temporal index

cisaacstern:
Closes #169 

Reimplements ecoscope core logic linked in #169 because it's just a few lines of boilerplate pandas, so I figure it's not worth depending on upstream for that.

Merging to #168 to unblock that PR.

cc @Yun-Wu and @walljcg for visibility (Thanks for understanding the moving fast w/out review, so I can finish e2e patrols. Happy to field feedback via follow-up issues!)

### DIFF
--- a/ecoscope_workflows/tasks/transformation/_indexing.py
+++ b/ecoscope_workflows/tasks/transformation/_indexing.py
@@ -1,0 +1,81 @@
+from typing import Annotated, Literal
+
+from pydantic import Field
+
+from ecoscope_workflows.annotations import AnyDataFrame
+from ecoscope_workflows.decorators import task
+
+
+# TODO: Leverage this information to generate more readable documentation
+# for the `directive` parameter of `add_temporal_index` task. For now, this
+# is not used anywhere, but leaving it here for future reference.
+strftime_directives = {
+    "%a": "Locale's abbreviated weekday name.",
+    "%A": "Locale's full weekday name.",
+    "%b": "Locale's abbreviated month name.",
+    "%B": "Locale's full month name.",
+    "%c": "Locale's appropriate date and time representation.",
+    "%d": "Day of the month as a decimal number [01,31].",
+    "%f": "Microseconds as a decimal number [000000,999999].",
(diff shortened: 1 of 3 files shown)


ecoscope_workflows/tasks/transformation/_indexing.py
```diff
@@ -0,0 +1,81 @@
...
+    "%X",
+    "%y",
+    "%Y",
+    "%z",
+    "%%",
+]
+
+
+@task
+def add_temporal_index(
```
walljcg, line 66: Here, the return of the task is the dataframe with a new index applied - but how do we propagate the index values we just created to the next task and eventual mapvalues() function? I guess we know the name of the index we just set and that param will be known by the later tasks...?

ecoscope_workflows/tasks/transformation/_indexing.py
```diff
@@ -0,0 +1,81 @@
...
+from pydantic import Field
+
+from ecoscope_workflows.annotations import AnyDataFrame
+from ecoscope_workflows.decorators import task
+
+
+# TODO: Leverage this information to generate more readable documentation
+# for the `directive` parameter of `add_temporal_index` task. For now, this
+# is not used anywhere, but leaving it here for future reference.
+strftime_directives = {
```
walljcg, line 12: Thank-you for adding these. The directives don't lend themselves to readability on their own!

Issue #102, opened 2024-07-26
Support setting literal values in yaml spec `with` block


cisaacstern:
Dependencies:
- [ ] #90

_**From the original discussion**_:

@atmorling yes good observation. You are correct that the create map and create plot tasks are nearly identical, with the exception of setting `widget_type="map"` or `widget_type="plot"`. And therefore could be combined into a single `create_html_widget_single_view` as long as `widget_type` (with options "map" or "plot") was also a parameter. If this ultimately seems better, I'm open to considering this. The reason I've kept them separate (at the cost of offending DRY sensibilities 😅 ) is to keep the compilation spec a little more descriptive, i.e.

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/4f63f8cef8b61f2cf9e473ce426b3aba0534d449/examples/compilation-specs/time-density.yaml#L11-L16

This ensures that the widget resulting from `create_map_widget_single_view` will have the correct value of `widget_type="map"` (since it's hardcoded into that task).

If we instead had 
```yaml
draw_ecomap: 
  geodataframe: calculate_time_density 
persist_text: 
  text: draw_ecomap 
create_html_widget_single_view: 
  data: persist_text 
```

Then that would require the user to know that they would also have to set `widget_type="map"` as a parameter. Currently, that happens in a totally different place (the web form) which seemed very odd). But having gotten to this point in my explanation (and I truly didn't anticipate getting here, so thanks for the question!) ... I'm now realizing that once #90 is finished, we have a path to setting explicit/literal values (that is, not just results from other tasks) in the yaml spec directly, so we could then do something like:

```yaml
- name: Persist HTML Text for Map
  id: save_map_html
  task: persist_text 
  with:
    text: ${{ ecomap.return }}
- name: Create Map Widget
  id: create_map_widget
  task: create_html_widget_single_view 
  with:
    widget_type: "map"
    data: ${{ save_map_html.return }}
```
Most of this ☝️ (`name`/`id`/`task`/`with` field names, etc.) is already encoded in #90. The new thing here is the idea that a literal value (`"map"`) could also be set in the spect itself. I've been wondering if we would encounter situations where we'd want to do this, but just hadn't run into one yet. This seems like a good motivator for that. What do you think? I'll open an issue to track this.

_Originally posted by @cisaacstern in https://github.com/wildlife-dynamics/ecoscope-workflows/pull/91#discussion_r1693235287_

atmorling:
[...]

PR #176, opened 2024-08-09
Add grouped single value widgets to dashboard

### LINKED ISSUE #119: Patrols Workflow: Add summary stats single value widgets to dashboard

cisaacstern:
closes #119 

Some notes for reviewers:

- I changed the single value tasks so that rather than providing the aggregation mode passed as a literal string, there is just a separate task per aggregation mode. This is for basically the same reason as discussed for widget factory tasks in #102; namely the spec doesn't currently support passing literal values (which I'd not been tracking when I reviewed the original single value widgets PR).
- I added an `apply_arithmetic_operation` task as a quick workaround to handle "unit conversion" but we may want to handle that in a more sophisticated way eventually. 🍺 Anyone care for a [pint](https://pint.readthedocs.io/en/stable/getting/overview.html)?
- Speaking of units: https://github.com/wildlife-dynamics/ecoscope-workflows/issues/177

### DIFF
--- a/ecoscope_workflows/tasks/analysis/__init__.py
+++ b/ecoscope_workflows/tasks/analysis/__init__.py
@@ -1,4 +1,5 @@
 from ._aggregation import (
+    apply_arithmetic_operation,
     dataframe_column_mean,
     dataframe_column_nunique,
     dataframe_column_sum,
@@ -8,6 +9,7 @@ from ._time_density import TimeDensityReturnGDFSchema, calculate_time_density
 
 __all__ = [
     "TimeDensityReturnGDFSchema",
+    "apply_arithmetic_operation",
     "dataframe_column_mean",
     "dataframe_column_nunique",
     "dataframe_column_sum",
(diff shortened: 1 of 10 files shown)


atmorling:
LGTM  - It might make sense in future to have the plotting tasks require the use of these agg functions, rather than handle it themselves internally.

ecoscope_workflows/tasks/analysis/__init__.py
```diff
@@ -1,8 +1,22 @@
-from ._aggregation import aggregate
+from ._aggregation import (
+    apply_arithmetic_operation,
+    dataframe_column_mean,
```
walljcg, line 3: Since dataframe columns are pd.Series shall call these 'series_mean' etc to shorten?

cisaacstern:
> It might make sense in future to have the plotting tasks require the use of these agg functions, rather than handle it themselves internally.

@atmorling can you expand a bit further on what you mean by this? Either here or in a new issue. 🙏

PR #179, opened 2024-08-13
Lithops executor

### LINKED ISSUE #116: LithopsExecutor
Following #90 we can add a lithops template.

cisaacstern:
closes #116


cisaacstern:
Ok so I've been trying to figure out the way the async graph scripts need to look and here is a runnable toy example using only built-in Python (no ecoscope-workflows) which I think accurately captures the spirit:

```python
from dataclasses import dataclass
from graphlib import TopologicalSorter
from typing import Any, Callable


@dataclass
class Node:
    async_callable: Callable
    params: dict[str, Any]

@dataclass
class Future:
    future: Any

    def gather(self):
        # in a true async environment, this would actually
        # gather a result from the future. For now, we just
        # return the future, which is the same as the result,
        # because this is a toy example.
        return self.future

@dataclass
class Result:
    result: Any

@dataclass
class Depends:
    node: str


def inc(x: int) -> int:
    return x + 1

def dec(x: int) -> int:
    return x - 1

def add(x: int, y: int) -> int:
    return x + y


nodes: dict[str, Node] = {
    'A': Node(inc, {"x": 1}),
    'B': Node(dec, {"x": Depends("A")}),
    'C': Node(add, {"x": Depends("A"), "y": 1}),
    'D': Node(add, {"x": Depends("B"), "y": Depends("C")}),
}
graph = {"D": {"B", "C"}, "C": {"A"}, "B": {"A"}}
results: dict[str, Result] = {}

ts = TopologicalSorter(graph)
ts.prepare()
while ts.is_active():
    ready: list[str] = [name for name in ts.get_ready()]
    print(f"Ready: {ready}")
    futures: dict[str, Future] = {}
    for name in ready:
        node = nodes[name]
        hydraded_params = {}
        for k, v in node.params.items():
            if isinstance(v, Depends):
                v = results[v.node].result
            hydraded_params[k] = v
        future = node.async_callable(**hydraded_params)
        print(f"  Started {name}")
        futures[name] = Future(future)
    for name in ready:
        results[name] = Result(futures[name].gather())
        print(f"  Finished {name} with result {results[name].result}")
        ts.done(name)

```
I'll work on refining this now for our actual use case.

cisaacstern:
The point address by the of above toy example is that at any given moment, our maximum task-level parallelism (this does not include intra-task parallelism, i.e. map/mapvalues), is constrained by the nodes which are released as "independent" at any given layer of [`graphlib.TopologicalSorter.get_ready`](https://docs.python.org/3/library/graphlib.html#graphlib.TopologicalSorter).

For our Patrol Workflow example, those layers can be generated (as task instance 'id' strings), like so:

```python
import pathlib
from graphlib import TopologicalSorter

import yaml

from ecoscope_workflows.compiler import Spec

examples = pathlib.Path(__file__).parent.parent / 'examples'

with open(examples / "compilation-specs" / "patrol_workflow.yaml") as f:
    spec = Spec(**yaml.safe_load(f))

ts = TopologicalSorter(spec.task_instance_dependencies)
ts.prepare()
while ts.is_active():
    ready = [node for node in ts.get_ready()]
    print(ready)
    for n in ready:
        ts.done(n)
```
And look like this:
```
['groupers', 'patrol_obs', 'patrol_events']
['patrol_reloc', 'filter_patrol_events']
['patrol_traj', 'pe_add_temporal_index', 'patrol_events_bar_chart', 'patrol_events_pie_chart']
['traj_add_temporal_index', 'td', 'split_pe_groups', 'patrol_events_bar_chart_html_url', 'patrol_events_pie_chart_html_url']
['split_patrol_traj_groups', 'td_map_layer', 'patrol_events_map_layers', 'patrol_events_bar_chart_widget', 'patrol_events_pie_chart_widget']
['patrol_traj_map_layers', 'total_patrols', 'total_patrol_time', 'total_patrol_dist', 'avg_speed', 'max_speed', 'td_ecomap']
['combined_traj_and_pe_map_layers', 'total_patrols_sv_widgets', 'total_patrol_time_converted', 'total_patrol_dist_converted', 'avg_speed_sv_widgets', 'max_speed_sv_widgets', 'td_ecomap_html_url']
['traj_patrol_events_ecomap', 'total_patrols_grouped_sv_widget', 'total_patrol_time_sv_widgets', 'total_patrol_dist_sv_widgets', 'avg_speed_grouped_widget', 'max_speed_grouped_widget', 'td_map_widget']
['traj_pe_ecomap_html_urls', 'patrol_time_grouped_widget', 'patrol_dist_grouped_widget']
['traj_pe_map_widgets_single_views']
['traj_pe_grouped_map_widget']
['patrol_dashboard']
```
So the next step is to combine the toy example (with refinements) with this code, to actually create a runnable async script.

cisaacstern:
At this point, I am able to execute trivial DAGs with Lithops (e.g. integer arithmetic):

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/e196718c674aff460c74ccd05414b028afe1e034/tests/test_graph.py#L258-L290

But for some reason the executor hangs on non-trivial workloads (passing around dataframes, for example), such as I have been attempting in his (very messy, WIP) [script](https://github.com/wildlife-dynamics/ecoscope-workflows/blob/lithops-executor/examples/dags/mode_map_dag.script_async.py).

Lithops clearly can handle doing this type of work, as proven in #45, so now it's just a matter of connecting the golden spike between the graph executor here, and the logic in #45.

I assume the slowness I'm seeing has something to do with serializing larger and more complex objects (such as dataframes) and/or the script's module/namespace correctly into the local executor... Presumably carefully stepping through and comparison with #45 will reveal the path forward here.

cisaacstern:
The current prototype async script runs e2e for me locally with config

```yaml
# .ecoscope-workflows-tmp/.lithops-config.yaml

lithops:
  backend: localhost
  storage: localhost
  log_level: DEBUG
  data_limit: 8
```

and command

```console
$ LITHOPS_CONFIG_FILE=.ecoscope-workflows-tmp/.lithops-config.yaml \             
ECOSCOPE_WORKFLOWS_RESULTS=./examples/output \
python3 examples/dags/mode_map_dag.script_async.py --config-file examples/params/mode_map_params.yaml
```

Note that I for some reason had a broken lithops local runtime in a prior conda environment, resulting in mysterious hanging while some process in the multiprocessing pool was waiting on some `lock.aquire()` ... anyway, this was resolved by just creating a new venv with Python's builtin `venv` module, and reinstalling workflows + lithops in it. (I'm not sure that the problem was necessarily specific to conda, but I happened to solve it without conda.)

Next step is to get the async script to actually compile, so that means a new async graph template. (There could be a sync graph template, for debugging, as well?) For the async graph at least, the executor and executor config could be set in configuration perhaps (rather than having a specific "lithops-graph" template).

cisaacstern:
In a textbook case of karmic retribution for over-confidence, I excitedly claimed to @walljcg  about 12 hours ago that I was “cruising” to the finish line of this PR.

Appropriately, then, the last 12 hrs have not, in fact, been smooth sailing, but rather an experience of being absolutely humbled by the initially seemingly incredible challenge of building a functional environment that can both run lithops, and also maintain the geopandas and numpy upper version bounds required for ecoscope.

Up until now, I have (somehow :shrug:) been able to avoid using mamba/conda for local or CI builds, relying only on pip + pyproject.toml, which has worked fine (even for ecoscope + geopandas). No longer! As reflected in the recent commit history, the only way I was able to get a working environment was to migrate to mamba/conda. Perhaps this tech debt was going to come due sooner or later (since geopandas does, of course, recommend this), but I had been holding out hope that the considerably lighter-weight pip/pyproject.toml setup would be workable. I no longer see that as a viable path, and it is probably for the best in the long term that we get on better footing as regards environment management in this repo.

The good news:
- Using the env files added here, this _does_ run locally for me
- I believe I am very close to getting this to run in CI (which as the commit history attests, has been a learning process)
- Depending on how quickly I can resolve the CI tomorrow AM, I believe getting this PR ready for review by Wednesday EOD is still attainable as a reach goal.

cisaacstern:
With tests passing, I'm now focusing on getting the async script example (which I wrote manually) to compile from a template, and then I can test it as part of the test_examples::test_end_to_end as an alternate compilation parametrization.

cisaacstern:
We can now compile async graphs with a template. Next steps:

- [ ] Implement mapvalues on AsyncTask
- [ ] Defer dependencies in partial somehow

cisaacstern:
Test for local multiprocessing are passing. An initial glimpse of performance. For the _very_ small scale of data we are using in test (KB or low single-digit MB, I believe), the sequential script is considerably faster than the lithops multiprocessing script: 10 secs for sequential vs 57 for lithops multiprocessing. Keep in mind lithops has had _zero_ optimization at this point, and I believe there is a lot of room to improve that. Also, it is very reasonable to expect at small data scales that multiprocessing overhead (regardless of the executor) will outweigh performance benefit. But still, a good first reference.

Screenshot 2024-08-21 at 11 13 05 PM

cisaacstern:
To navigate to the API docs build for this PR, click the readthedocs `Details` link on the checks window:

Screenshot 2024-08-22 at 12 24 39 AM



...[or just click the same link here](https://ecoscope-workflows--179.org.readthedocs.build/en/179/ecoscope_workflows/index.html). This is good place to begin familiarizing yourself with the code.

Of note:

- There are now both [`SyncTask`s and `AsyncTask`s
](https://ecoscope-workflows--179.org.readthedocs.build/en/179/ecoscope_workflows/decorators.html)
- `AsyncTask`s are executed with an instance of an implementation of a [`AsyncExecutor`](https://ecoscope-workflows--179.org.readthedocs.build/en/179/ecoscope_workflows/executors/base.html#ecoscope_workflows.executors.base.AsyncExecutor). They have all the same methods as `SyncTask`s, except they return Futures, rather than directly returning values. This lets us call them asynchronously and get a future back, rather than waiting around for them to finish running before we call other tasks.
- The [`LithopsExecutor`](https://ecoscope-workflows--179.org.readthedocs.build/en/179/ecoscope_workflows/executors/lithops.html#ecoscope_workflows.executors.lithops.LithopsExecutor) is our first implementation of `AsyncExecutor`. In the future, any API that allows us to execute python callables and get futures in return, could theoretically be implemented as well.
- All `@task`s are `SyncTask`s by default, but they can be transformed into `AsyncTask`s by re-assigning their executor to be an asynchronous one, e.g.:
   https://github.com/wildlife-dynamics/ecoscope-workflows/blob/5b610372a3dececce2a00ce2fb756c29fb8652dc/examples/dags/patrol_workflow_dag.script_async.py#L115
- The compiler now knows how to translate our standard compilation specs into a DAG of async tasks, which are executed by a [`Graph`](https://ecoscope-workflows--179.org.readthedocs.build/en/179/ecoscope_workflows/graph.html#ecoscope_workflows.graph.Graph) , which functions as a lightweight scheduler, topologically sorting the DAG, and releasing batches of work as they become ready (i.e., once all of their dependencies are resolvable from futures): 

    https://github.com/wildlife-dynamics/ecoscope-workflows/blob/5b610372a3dececce2a00ce2fb756c29fb8652dc/ecoscope_workflows/graph.py#L78-L85
- All of our pre-existing compilations specs are tested in CI in this execution mode now, and all those tests are passing. Some notes on performance in https://github.com/wildlife-dynamics/ecoscope-workflows/pull/179#issuecomment-2303868051.

.github/workflows/unit-test.yaml
```diff
@@ -12,43 +12,47 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
+        environment-name: [
+          "py3.10-ecoscope1.8.2",
```
Yun-Wu, line 16: Do we want to attach ecoscope version here? It means every time we upgrade ecoscope we need to change the github action too

cisaacstern (commented):
Just reread through the whole PR, going to merge.

PR #186, opened 2024-08-16
Speed up tests: drop gsm dep, pytest-xdist, cache venv

### LINKED ISSUE #185: Use pytest-xdist
Our tests already run pretty fast, but this might make them faster

### LINKED ISSUE #187: Drop google-secrets-manager from test deps
We will be handling this on the server so we don't need this here, and it's slowing down our test runs.

### LINKED ISSUE #188: Better dependency caching in CI
The actions/setup-python cache doesn't save that much time.

Let's try it like this instead: https://github.com/actions/setup-python/issues/881#issuecomment-2147747696

cisaacstern:
closes #185 
closes #187
closes #188


cisaacstern:
Just a little tests maintenance to keep developer morale high

cisaacstern:
Average pytest run reduced from ~2mins -> ~40s. 🎉

PR #189, opened 2024-08-19
Add additional style parameters

Yun-Wu:
- **add extra style fields**
- **update json**


cisaacstern (approved):
Love it! Thanks @Yun-Wu

PR #205, opened 2024-08-23
Micromamba for ci

### LINKED ISSUE #206: Tests are broken due to ecoscope core incompatibility with numpy>=2
Tests are broken.

I will address this in #205

cisaacstern:
closes #206

### DIFF
--- a/.github/workflows/unit-test.yaml
+++ b/.github/workflows/unit-test.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: Setup Micromamba
         uses: mamba-org/setup-micromamba@v1
         with:
-          environment-file: ./runtimes/${{ matrix.environment-name }}.yml
+          environment-file: ./ci/envs/${{ matrix.environment-name }}.yml
           cache-environment: true
           cache-downloads: true
           init-shell: bash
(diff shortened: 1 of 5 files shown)


cisaacstern:
Thanks @Yun-Wu for your feedback in https://github.com/wildlife-dynamics/ecoscope-workflows/pull/179#discussion_r1728605009, which is relevant here too, is this is the same implementation.

I have starting thinking through some alternate pathways in #207, but those will require more thought, so if its helpful to just use this PR's diff to unblock other work for now, please feel free to merge this (or not, if you are not blocked).

.github/workflows/unit-test.yaml
```diff
@@ -12,43 +12,47 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
+        environment-name: [
+          "py3.10-ecoscope1.8.2",
+          "py3.11-ecoscope1.8.2",
+          "py3.12-ecoscope1.8.2",
+        ]
```
Yun-Wu, line 19: Do we want to attach ecoscope version here? It means every time we upgrade ecoscope we need to change the github action too

Yun-Wu: Synced offline and will resolve it later.

Yun-Wu (approved):

PR #210, opened 2024-08-26
Make pie chart grouped in patrols workflow

atmorling:
Also adds some extra data to the `get-patrol-events` example parquet so the date range aligns with `get-patrol-observations` 
xref #211


cisaacstern (approved):
Sorry for the delay on this! Looks great!

We just need to recompile one file to resolve conflict then we can merge.

PR #212, opened 2024-08-26
Try dropping tools.setuptools py-modules setting from pyproject.toml

cisaacstern:
An attempt to fix some install issues.


cisaacstern:
For those seeing this and wondering what it was for (my future self included): I was having an issue where installing this repo in a conda/mamba repo with pip would result in ModuleNotFound errors on attempts to import it. This turned out to be the culprit.

PR #214, opened 2024-08-28
Build worker container with custom conda channel

cisaacstern:
This is an experiment to see if this approach will allow us to build smaller worker containers for deployment.

The basic concept is that:

- Our [worker prototype container](https://github.com/cisaacstern/patrol-workflow-build/blob/main/build/Dockerfile), as built from the CI envs in this repo, is unmanageably large (in the range of 3GB)
- @daniefdz has helped me analyze this and we have concluded that the source of almost all of the bloat in this container are python dependencies. Indeed, I took a look at disk usage in a shell on the container and the `/env/python/lib/` directory for our conda env is ~2GB!
- Currently many of our dependencies are installed with pip, because they are not available on conda-compatible channels
- Conda/Mamba/Micromamba environment resolvers can potentially reduce duplication of dependencies if the packages installed by them are hosted on conda-compatible channels (which pip is not)
- Long-term, the public `conda-forge` channel is a possibility, but that channel does not move fast enough for our pace of iteration (I have hosted packages there before), and is better suited IMHO to distribution of packages to _users_, not necessarily for hosting packages for our own deployment needs.
- An alternative for us is to host our own "build channel" which can also just be a local directory. The artifacts in that channel must be build using some conda-compatible build system. The most performant/contemporary one I've identified is [rattler-build](https://prefix-dev.github.io/rattler-build/latest/)
- To make this work, we need to:
   1. User rattler-build to build conda-compatible packages for all of our dependencies which do not already offer them. This needs to be done from bottom -> up, in terms of dependency resolution, because lower level packages are used for subsequent builds of higher-order ones. (E.g. the local build of `ps-mem` is used to build `lithops`, the local build of our `lonboard` fork is used to build `ecoscope`, etc.)
   2. Define an `environment.yml` that is similar to the one we are currently using, except which pulls all of these dependencies from the local channel (rather than pip)
   3. This _may_ allow us to share binary dependencies more efficiently between packages in our environment, thus reducing bloat. It will also provide us a sound basis upon which to customize builds (for example... we could build GDAL in this pipeline _without_ NetCDF support, because we are not using that).

So far, the tree of custom packages I have come up with looks like this:

```
psmem        er_client   lonboard (fork)
  |                 \    /
lithops            ecoscope
    \               /
    ecoscope-workflows
```


cisaacstern:
The core approach here definitely works from a technical standpoint, as partially shown in CI here already, as well as further in my local testing. I just need to get to the bottom of this tree of packages (which is not that long) to get a sense of if it actually has a container size benefit for us. Will report back soon!

cisaacstern:
Ok now things are getting interesting. Using `matplotlib-base` saves 0.85 GB:

```
REPOSITORY                 TAG             IMAGE ID       CREATED          SIZE
464a19d                    latest          7394eb8373f4   2 minutes ago    3.14GB
791d2b1                    latest          6bbb7b5f09bf   45 minutes ago   3.95GB
```

cisaacstern:
Status update: from an original **`3.95GB`**, we are no down to 🎉 **`1.58GB`** 🎉, so in ~2 days of work I've cut `2.37GB` from the worker image, without any loss in functionality (end to end examples tests pass inside the container, in CI, as noted in the checks on this PR). Here's a little glimpse of the evolution of that process, from my local machine: 

```console
$ docker images
REPOSITORY                                         TAG             IMAGE ID       CREATED             SIZE
us.gcr.io/ecoscope-poc/ecosope-workflows-runtime   latest          7b30e23b0474   17 minutes ago      1.58GB
ecoscope-workflows-worker                          latest          231beb37e429   About an hour ago   1.55GB
f601666                                            latest          c98dbf529368   3 hours ago         1.83GB
fd9506e                                            latest          3ca580ba5419   17 hours ago        1.84GB
464a19d                                            latest          49addda33fb6   19 hours ago        3.14GB
791d2b1                                            latest          6bbb7b5f09bf   20 hours ago        3.95GB
```

The `us.gcr.io/ecoscope-poc/ecosope-workflows-runtime` was actually built and pushed to GCP with `lithops build`, so we have every reason to believe this will run on Cloud Run, however that has not been tested yet. Next steps:

- [ ] Move https://github.com/wildlife-dynamics/ecoscope-server/pull/71 into a PR in workflows ontop of this, so we can actually integration test this final build with lithops on Cloud Run
- [ ] Figure out how we want to integrate the CI here into our actual workflows deployment process, which probably is linked to https://github.com/wildlife-dynamics/ecoscope-workflows/issues/216 somehow (that is, tying container builds to actual commits to a workflows `build` directory)
- [ ] Clean up and document all of this


P.S. - I think we've possibly the hit the point of diminishing returns on optimizing this image without major refactors to ecoscope core. With some more effort we might be able to eek a 1-2 hundred more megabytes of savings out of further cleaning up the env, perhaps by following some more of the suggestions in [this blog](https://uwekorn.com/2021/03/01/deploying-conda-environments-in-docker-how-to-do-it-right.html). To get smaller than that, I think we would need to rethink some core assumptions (e.g., maybe don't depend on scipy, and use some other compiled routine for the work scipy is doing for us, etc.).

cisaacstern:
Note on next steps:

- [ ] Currently working on first item from previous comment checklist; i.e. testing on cloud run from the rattler build workflow
- [ ] After that, it seems the clearest path forward to integrate the script runner is to use the same base image as the lithops proxy, but with a different entrypoint (the script runner, rather than the lithops proxy) ... so this means maybe factoring those steps into a separate image file and making a worker image that inherits from it (which looks like the current image) as well as a "client/script-runner" image which runs the fastapi script runner microservice

cisaacstern:
Notes to self for tomorrow:
- The current issue is https://github.com/moby/moby/issues/5509
- The most obvious solution here is to not use distroless so we can run a shell. That does cost us a few (maybe up to 4) hundred MiB in image size, so it's worth considering if there is another approach, but there may not be.

cisaacstern:
Exciting to see tests pass. On further reflection perhaps simply running in debian-slim or similar is sufficient at our scale; the lack of shell on distroless certainly causes some headaches.

cisaacstern:
Tues EOD notes:
- Building from lockfile is much faster.
- Would probably/presumably be even moreso if we could mount a download cache to build.
- The `lock.txt` committed here is not tractable as a the final approach to locking builds, so need to incorporate a process into the build sequence by which this file is generated, and associated with a hash key tied to the sum of the `rattler-build` recipe files. Then we can fail out and/or re-lock if the recipes have evolved since the last lock.

cisaacstern:
Todo: allow disabling `Build` via PR tag, but require that it does pass before merging.

cisaacstern:
Moving pytest to `only-deps` image seems to have saved about ~20-30 seconds from the base image build. Now will try using `--freeze-installed` instead of `--no-deps` in base, which is almost certainly safer, and maybe even faster as well.

cisaacstern:
maybe undue special-casing of mamba envs 🤔 to fix unit test failures

cisaacstern:
Without `test-docker` label:

Screenshot 2024-09-05 at 5 10 14 PM

Skipped tests are enabled with addition of `test-docker` label.

cisaacstern:
For the scale of this PR, I concede that it is not as well documented (in prose, that is) as it could be, but I hope the clarity of the implementation makes up for that shortcoming.

In terms of what the implementation is, here are the main points:
- Testing is now bifurcated into two ✌️ separate streams of work. 
- The first stream is based on a standard pip installation of `ecoscope-workflows` and covers all tests not marked as `pytest.mark.requires_ecoscope_core`. This covers the fundamental decorator tests, workflows compilation and execution logic, as well as all tasks which do not depend on `ecoscope` core. These tests run very quickly (~30 seconds) and can give the developer a fast indicator if something fundamental is broken, as well as serve as a helpful reference point for the type of work that is possible based on a pip-only installation.
- The second stream (the `Build` tests) is more time-consuming but more comprehensive. In this stream, we build a conda package for `ecoscope-workflows`, which we then use to create a test (micromamba) environment which includes `ecoscope` core, which allows us to run all the tests (including the end-to-end workflows tests).
- If the `test-docker` label is applied to the PR, the conda package for `ecoscope-workflows` is also used to build a docker container, in which the end to end tests are also run, both locally (i.e. locally within the GitHub Actions runner), as well as remotely (i.e. by deploying to GCP Cloud Run).
- The container artifacts which are used for the docker tests are published here: https://github.com/orgs/wildlife-dynamics/packages?repo_name=ecoscope-workflows (currently private, lmk if you do not have access). This is useful for caching (many times they can be reused), and also for debugging (if something fails in CI, you can pull the same image which was used and run/test it locally).
- As for the conda package for `ecoscope-workflows`, that is also downloadable for each workflow run, it is the Artifact labeled `rattler-artifacts` at the bottom of the workflow run summary page, e.g.: https://github.com/wildlife-dynamics/ecoscope-workflows/actions/runs/10734707742
- This `rattler-artifacts` Artifact is a directory containing a custom conda channel with our `ecoscope-workflows` package. You can download it and use it to install into a conda environment like so (where `file://$(pwd)/tmp/conda-channel` is the unzipped artifact):
   https://github.com/wildlife-dynamics/ecoscope-workflows/blob/a7a4f2c783d6390a774247fc56ee69def06ad674/.github/workflows/test-build.yaml#L124-L128
- Eventually we can establish a release process where by we publish this artifacts to `conda-forge` or possibly https://prefix.dev/channels, but for now the local directory works. Note that the directory contains recipes for many other packages in addition `ecoscope-workflows`, those are our dependencies which do not otherwise offer a conda package release on a public registry (that I am aware of). And even if they did, it's possible we'd still want to manage (at least some of) them ourselves, so we can fine tune the transitive dependencies they bring in (i.e. opting for slimmer dependency sets than the official release might otherwise offer).

Note that there is a bit of noise in this PR because I ended up needing to bump pre-commit/mypy versions, which led to some linting changes that are not directly related to above.

Next steps:

- Wire in the script runner. This PR just got too big and I wasn't able to include that here. This will require its own Dockerfile, which will share a base with the existing images in this PR. The build/test process for this additional container can be attached into the `Build` workflow here as an additional job / testing pathway.
- Define a separate workflow for building and pushing container image _releases_. Notably the container builds here very-close-to production containers, but are not exactly the same as what we will want to deploy to production. For example, the version of `ecoscope-workflows` included in them is just from the head ref of the PR, and not from a release. So come to think of this, this is probably an `on: release` workflow. Which can push an official conda package release to a public registry, then use that conda release to build and tag an associated worker container. (🤔 As I am thinking of it, these will be workflow-specific, so I can think about how to approach that as part of the script-runner work.)
- Benchmark import and runtime speed of the `ecoscope-workflows` install in the conda environment. I have the general impression that it's a little laggy, but I'm not positive if that's the case. If so, look into if/how that lag might be improved.

cisaacstern:
To any prospective reviewers here ... as I mentioned to @walljcg earlier today... I am exploring https://pixi.sh/latest/ as an alternative to micromamba and at the risk of still not being done with this ... this is seemingly like an INCREDIBLY good alternative (faster, more reproducible, simpler). The fundamental structure here can remain somewhat similar, but we can probably reduce effort, build time, and complexity by using `pixi.sh`. I will follow up with updates.

cisaacstern:
Ok yes so clearly `pixi.sh` is the way. TL;DR we can eliminate the entire step of caching a base image, because the solve from scratch is as fast as (or perhaps 30 seconds faster than) the micromamba solve building on the base.

Almost done with that over in https://github.com/wildlife-dynamics/ecoscope-workflows/pull/220, can finish Monday.

Reviewers can hold off until that is done, I will re-ping here 😄

cisaacstern:
Superseded by #265

PR #219, opened 2024-09-03
Swap the order of Path and Url in type alias .

ANAMASGARD:
The type alias checks whether the input is Path first .


cisaacstern:
@ANAMASGARD thanks for picking up this issue. We may end up taking a different solution to this than the one described in the linked issue, so I am going to mark this as draft while I consider that, to signal that we are still deliberating on the best path forward.

In the meantime, may I ask how you found our repo, and what prompted you to contribute? We always appreciate new contributors and it helps to know how they have found us.

Issue #217, opened 2024-09-02
Fix README inaccuracies


cisaacstern:
Some things in the readme are stale/out-of-sync with the latest code. In particular, I noticed that the compilation spec examples are not reflective of the latest spec.

cisaacstern:
Per https://github.com/wildlife-dynamics/ecoscope-workflows/pull/225#discussion_r1769696336 pixi installation and usage must now also be added as part of this work.

PR #225, opened 2024-09-12
Pixi build directories for workflows

### LINKED ISSUE #216: `ecoscope-workflows compile` command should generate a build directory (not just a single artifact)

cisaacstern:
Builds (no pun intended) on top of #222

closes #216


cisaacstern:
TODO:
- [ ] some other basic setup.py or equivalent to install ecoscope_workflows from local path

cisaacstern:
TODO:
- [ ] regenerate mode-map build dir via a `compile` command (see WIP data modelling)
- [ ] generate and test remaining examples this way

cisaacstern:
Quite close to generating the PixiToml here, the remaining work on that involves:
- [ ] A more robust and stable way of ensuring that the dependencies/`Features` we will use to compose the dependencies for this generated file are synchronized with pyproject.toml. This should encompass both task-specific Features as well as the default dependencies. Currently I have repeated the values of the former from pyproject.toml in `task/features.py`. This is brittle an can easily fall out of sync. Instead, I think a better idea (which can work for both categories of dependency) is:
    - Change to hatchling for build system (because it has better support for build hooks than setuptools)
    - Create a hatchling [build hook](https://hatch.pypa.io/1.12/plugins/build-hook/reference/) that injects a module, `_features.py` at build/install time. This module can contain a dictionary of all Features/dependencies present in pyproject.toml at time of build. Like `_version.py`, this module would be gitignored, so that it is only generated per-install.
    - A user-facing `features_registry` in `registry.py` could contain a read-only MappingProxy pointing to the generated `_features.py`, and could be used like:
    
    ```python
    from ecoscope_workflows.registry import features

    @task(requires=[features["ecoscope-core"]])
    def f(...): ...
    ```

    and perhaps there would actually be a second `dependencies` proxy mapped to a generated `_dependencies.py` (or both dependencies and features could be mapped to dictionaries in `_pixi.py`) ... either way, the `dependencies` could be used as the basis for the default PixiToml.dependencies ... so essentailly, PixiToml.dependencies is assigned to be the union of these default dependencies + the dependencies of all features requested by all tasks in the workflow. ... (🤔 not sure if we want to extract deps from features, or leave them as-is and make a "runtime environment" ... because what we don't want is too loose other non-dependency information from that features; could always redo this later though).
- [ ] Finally, and mostly unrelated to above, we need to generate the symlink to the installation of ecoscope workflows and set that as an editable pypi-dependency, along with (i believe) a generated minimal pyproject.toml, as we cannot assume we'll have access to the original one if we're not in a development context.

cisaacstern:
Restating some offline comments, so they are captured in the public record here:

Following various false starts and design conversations over the last week, I’m finally gaining traction with an implementation here that feels workable. This implementation revives an idea which I’d previously explored and set aside, namely: namespace packages, [as demonstrated in titler](https://github.com/developmentseed/titiler/tree/main?tab=readme-ov-file#packages). This separates workflows into two separate packages: `ecoscope-workflows.core` (with no `ecoscope` dependency) and `ecoscope-workflows.ext.ecoscope` (which introduces the `ecoscope` dependency). I’d resisted this approach previously because it makes the repo structure a bit more Byzantine than I’d prefer, but the more I experimented with it, I came to see that the extensibility we are seeking in workflows is truly best served by this style, insofar as it facilitates the ability to cleanly delineate core dependencies from extension task dependencies, and sets us up with a much clearer onramp for supporting user-defined task extensions. Most importantly for our near-term concerns, it represents a side-step around the issue that has been a design thorn in our (my) side for the last week: task-specific requirements. In this paradigm, we don’t need to attach that metadata at the @task level, but rather can just specify additional dependencies to come in with the extension package(s) in normal Python ways. Happy to review this path further with anyone interested tomorrow. :pray:

cisaacstern:
Next steps:

- [ ] Add an `extensions:` field to the compilation spec and use it in mode-map to inform generation of a `pixi.toml` that specifies both `ecoscope-workflows.core` and `ecoscope-workflows.ext.ecoscope`. Sphinx offers a good [design reference](https://github.com/pangeo-forge/pangeo-forge-runner/blob/8310621af4888ca8c2b16249b5ad1c95c7635dd8/docs/conf.py#L16-L17) [here](https://www.sphinx-doc.org/en/master/usage/extensions/index.html).

cisaacstern:
I just realized thanks to @juanlescano-ng that I need to .gitignore that `.conda` directory. That's not supposed to be here. 😅

cisaacstern:
next steps:
- [ ] finish pixi.toml as part of workflow artifacts, and dump it with `compile`
- [ ] add a "lock" step to generate the lockfile after compile
- [ ] migrate all compiled dags to build dirs

cisaacstern:
Following `py-rattler` example, it's probably most legible if we split pyproject.toml and pixi.toml configs for our packages, xref https://github.com/conda/rattler/tree/main/py-rattler

.gitignore
```diff
@@ -1,5 +1,14 @@
...
 
+# custom build hook / feature registry logic
+ecoscope_workflows/_registry.py
+
+# pixi
+.pixi
+
+# vendor
+vendor/artifacts
+
```
cisaacstern, line 11:
```suggestion
# pixi
.pixi
```

src/ecoscope-workflows-core/ecoscope_workflows_core/registry.py
```diff
@@ -54,6 +53,9 @@ def recurse_into_tasks(
                 anchor=module.__name__,
                 tags=obj.tags or [],
             )
+        elif module.__name__.endswith(".features"):
+            # reserved module name, skip
+            continue
```
cisaacstern, line 58:
```suggestion
```

publish/version-check.py
cisaacstern, line 1: - [ ] Either fix this as part of this PR or remove it and defer it as future work

requirements-core.toml
cisaacstern, line 1: - [ ] See comment above. If we do version-checking as part of this PR, this file is meant to be the ground truth against which other versions are compared.

src/ecoscope-workflows-core/README.md
cisaacstern, line 1: - [ ] This first attempt at describing installation procedure in the README has already gone a bit stale. Either remove it or fix it.

src/ecoscope-workflows-core/ecoscope_workflows_core/__main__.py
```diff
@@ -61,14 +96,17 @@ def get_params_command(args):
...
-    compilation_spec = Spec(**yaml.safe_load(args.spec))
+    compilation_spec = Spec(**yaml.load(args.spec))
     outpath = args.outpath
     write_png(compilation_spec, outpath)
 
 
 def connections_create_command(args):
+    known_connections = {}  # FIXME: fix this
+    if args.type not in known_connections:
+        raise ValueError(f"Unknown connection type '{args.type}'")  # FIXME: fix this
```
cisaacstern, line 109: - [ ] I had to remove `known_connections` from registry.py as part of the switch to the two-package system ... either reimplement it as an actual lazy registry, in the same vein as the tasks registry or find some other way to deal with this workaround.

If we want to configure connections from the CLI then we do need the registry. Alternatively, if we deprecate managing configuration ourselves (which is appealing from both a security and a maintenance standpoint), then we actually don't need the registry or this command on the CLI.

src/ecoscope-workflows-core/ecoscope_workflows_core/__main__.py
```diff
@@ -61,14 +96,17 @@ def get_params_command(args):
 
 
 def visualize(args):
-    from ecoscope_workflows.visualize import write_png
+    from ecoscope_workflows_core.visualize import write_png
 
-    compilation_spec = Spec(**yaml.safe_load(args.spec))
+    compilation_spec = Spec(**yaml.load(args.spec))
     outpath = args.outpath
     write_png(compilation_spec, outpath)
```
cisaacstern, line 103: - [ ] maybe we should put this into the WorkflowArtifacts, rather than it being a standalone command (as is the trend for the other outputs.

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/io/_persist.py
```diff
@@ -4,7 +4,8 @@
 from pydantic import Field
 from pydantic.json_schema import SkipJsonSchema
 
-from ecoscope_workflows.decorators import task
+from ecoscope_workflows_core.decorators import task
+from ecoscope_workflows_core.serde import _persist_text
```
cisaacstern, line 8: Move `serde._persist_text` implementation into `tasks.io.persist_task` module and remove `serde` as a top level module?

This is a holdover from an earlier design pattern where we might've been using `serde` features for serialization between tasks (which we are no longer doing, but I'm not sure if that may or may not be useful as an optimization?) 🤔

src/ecoscope-workflows-core/ecoscope_workflows_core/testing.py
```diff
@@ -0,0 +1,104 @@
...
+
+class Assertions(BaseModel):
+    result_stdout_contains: list[str]
+
+
+class TestCase(BaseModel):
+    name: str
+    description: str
+    params: dict
+    assertions: Assertions
```
cisaacstern, line 43: Move to `artifacts` and generate as part of compile

src/ecoscope-workflows-core/ecoscope_workflows_core/testing.py
```diff
@@ -0,0 +1,104 @@
+import contextlib
```
cisaacstern, line 1:
```suggestion
```

src/ecoscope-workflows-core/ecoscope_workflows_core/testing.py
```diff
@@ -0,0 +1,104 @@
...
+
+class TestCase(BaseModel):
+    name: str
+    description: str
+    params: dict
+    assertions: Assertions
+
+
+def test_case(script: Path, case_name: str, test_cases_yaml: Path, tmp_path: Path):
+    """Run a single test case for a workflow script."""
```
cisaacstern, line 47: move file io out of this function to make the signature more expressive

```diff
- test_cases_yaml: Path
+ test_case: TestCase
```

src/ecoscope-workflows-core/ecoscope_workflows_core/testing.py
```diff
@@ -0,0 +1,104 @@
...
+    stdout = proc.stdout.read().strip()
+    for expected_substring in test_case.assertions.result_stdout_contains:
+        assert expected_substring in stdout
+
+
+@contextlib.contextmanager
+def pixienv(path: Path):
+    sys.path.insert(0, path.as_posix())
+    yield
+    sys.path.remove(path.as_posix())
```
cisaacstern, line 104:
```suggestion
```

Unused, remove.

src/ecoscope-workflows-ext-ecoscope/README.md
```diff
@@ -0,0 +1 @@
+# `ecoscope-workflows.extension`
```
cisaacstern, line 1:
```suggestion
# `ecoscope-workflows-ext-ecoscope`
```

src/ecoscope-workflows-ext-ecoscope/ecoscope_workflows_ext_ecoscope/__init__.py
```diff
@@ -0,0 +1 @@
+from . import tasks as tasks
```
cisaacstern, line 1: Can this __init__.py be empty?

src/ecoscope-workflows-ext-ecoscope/ecoscope_workflows_ext_ecoscope/tasks/io/_earthranger.py
```diff
@@ -45,12 +39,15 @@ def get_subjectgroup_observations(
...
-        until=until,
+    return cast(
+        DataFrame[SubjectGroupObservationsGDFSchema],
+        client.get_subjectgroup_observations(
+            subject_group_name=subject_group_name,
+            include_subject_details=True,
+            include_inactive=include_inactive,
+            since=since,
+            until=until,
+        ),
```
cisaacstern, line 50: Note that this type of change is driven by the fact that we are actually running `mypy` in an environment with all of our project dependencies installed now.

src/ecoscope-workflows-ext-ecoscope/ecoscope_workflows_ext_ecoscope/tasks/results/_ecomap.py
```diff
@@ -191,8 +192,8 @@ def draw_ecomap(
     if len(legend_labels) > 0:
         m.add_legend(
             labels=legend_labels,
-            colors=legend_colors,
-            **(legend_style.model_dump(exclude_none=True)),
+            colors=[str(lc) for lc in legend_colors],
+            **(legend_style.model_dump(exclude_none=True)),  # type: ignore[union-attr]
```
cisaacstern, line 196: xref #244

tests/test_examples.py
```diff
@@ -58,19 +47,18 @@ def spec_fixture(request: pytest.FixtureRequest) -> SpecFixture:
...
+def test_generate_dag(spec_fixture: SpecFixture, dag_type: DagTypes):
     spec = Spec(**spec_fixture.spec)
     dag_compiler = DagCompiler(spec=spec)
-    dag_compiler.template = template
-    dag_str = dag_compiler.generate_dag()
-    script_fname = _spec_path_to_dag_fname(path=spec_fixture.path, template=template)
+    dag_str = dag_compiler.generate_dag(dag_type)
+    script_fname = _spec_path_to_dag_fname(path=spec_fixture.path, dag_type=dag_type)
     with open(EXAMPLES / "dags" / script_fname) as f:
         assert dag_str == f.read()
```
cisaacstern, line 63: - [ ] Account for this and the other "generate" tests in this module somehow before deleting it

src/ecoscope-workflows-ext-ecoscope/tests/test_compiler.py
```diff
@@ -21,6 +29,18 @@ def test_dag_compiler_from_spec():
...
+            - name: ecoscope-workflows-core
+              version: "*"
+              channel: file:///tmp/ecoscope-workflows/release/artifacts/
+          run:
+            - name: ecoscope-workflows-core
+              version: "*"
+              channel: file:///tmp/ecoscope-workflows/release/artifacts/
+            - name: ecoscope-workflows-ext-ecoscope
+              version: "*"
+              channel: file:///tmp/ecoscope-workflows/release/artifacts/
```
cisaacstern, line 43: - [ ] Try an inline table for these entries for line count /readability

src/ecoscope-workflows-ext-ecoscope/tests/test_config.py
cisaacstern, line 1: Related to comment on registry re: known_connections, perhaps we should just deprecate the entire config module

src/ecoscope-workflows-ext-ecoscope/tests/test_connections.py
```diff
@@ -83,7 +92,7 @@ def mock_toml_config(tmp_path: Path):
 
 
 def test_connection_named_from_toml(mock_toml_config: Path):
-    with patch("ecoscope_workflows.config.PATH", mock_toml_config):
+    with patch("ecoscope_workflows_core.config.PATH", mock_toml_config):
```
cisaacstern, line 95: If we deprecate the config module, then the only way to auto-magically resolve clients is env vars ... which should be ok?

cisaacstern (commented):
Comments including remaining punchlist items below follow a full read-through of this PR

requirements-core.toml
cisaacstern: Going to punt, tracking this in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/249

src/ecoscope-workflows-core/README.md
cisaacstern: Going to punt, tracking this in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/249

publish/version-check.py
cisaacstern: Going to punt, tracking this in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/249

src/ecoscope-workflows-core/ecoscope_workflows_core/__main__.py
cisaacstern: Going to punt on this but resolve through a quick follow-up, see https://github.com/wildlife-dynamics/ecoscope-workflows/issues/251

src/ecoscope-workflows-core/README.md
cisaacstern: Punting, see #217

src/ecoscope-workflows-core/ecoscope_workflows_core/__main__.py
cisaacstern: will resolve via https://github.com/wildlife-dynamics/ecoscope-workflows/issues/252

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/io/_persist.py
cisaacstern: tracking in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/253

src/ecoscope-workflows-core/ecoscope_workflows_core/testing.py
cisaacstern: will resolve as part of #254

src/ecoscope-workflows-core/ecoscope_workflows_core/testing.py
cisaacstern: xref #254

src/ecoscope-workflows-ext-ecoscope/ecoscope_workflows_ext_ecoscope/__init__.py
cisaacstern: Tests pass so not going to worry about this atm

src/ecoscope-workflows-ext-ecoscope/ecoscope_workflows_ext_ecoscope/tasks/io/_earthranger.py
cisaacstern: See #247 for additional context

src/ecoscope-workflows-ext-ecoscope/tests/test_config.py
cisaacstern: see #251

src/ecoscope-workflows-ext-ecoscope/tests/test_connections.py
cisaacstern: tracking in #251

src/ecoscope-workflows-ext-ecoscope/tests/test_compiler.py
cisaacstern: Punting and tracking in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/255

tests/test_examples.py
cisaacstern: Tracking in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/256

cisaacstern:
Punchlist done but why are tests not running 🤔

Issue #47, opened 2024-06-24
Generate YAML form from JSON Schema


cisaacstern:
Currently we generate fillable YAML forms as a separate code path from the JSON Schema.

This is problematic because the two representations can diverge, and also because it is easier to customize the JSON Schema representation (via Pydantic) than it is the YAML (which is just some ad hoc code I threw together myself).

An example of why this matters is @Yun-Wu's comment here https://github.com/wildlife-dynamics/ecoscope-workflows/pull/31#discussion_r1650204041, in which she observes that it would be nice to represent the form filler differently than the task function signature does. I agree! That kind of logic is relatively trivial with the Pydantic -> JSON Schema code path, but not supported for the (current) YAML code path.

I propose we punt on addressing your linked comment @Yun-Wu until we have resolved this issue, since it will be a more robust way of addressing this as opposed to continuing down the hacky path for YAML forms I have thus far been using.

h/t @JoshuaVulcan as this is relevant to your feedback re: JSON Schema as the source of truth 😄

PR #227, opened 2024-09-13
Task groups (for configuration UI)

### LINKED ISSUE #164: Task groups for logical organization of configuration wizard
Something like this: https://www.astronomer.io/docs/learn/task-groups

That allows us to programmatically represent what we are already doing with comments here:

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/98828f1662796eaad76c2f1925e38cc22579de0b/examples/compilation-specs/patrol_workflow.yaml#L3-L20

And that can included metadata such as a title + description for forthcoming UI requirements from @tiffanynwong.

Note that like Airflow Task Groups, these would be solely for UI intelligibility and not a logical/computational element.

cisaacstern:
closes #164


cisaacstern:
This PR adds the (optional) schema abstraction of `task-group`s. Like [Airflow task groups](https://www.astronomer.io/docs/learn/task-groups), these groups do not have any logical implications, rather they are used only for aesthetic/readability preferences of the workflow developer, and can also allow for such readability improvements in the UI configuration forms.

As shown in the patrol workflow schema in this PR, `task-group`s can be used in combination with regular task instances, which may be useful in the case where certain tasks are not part of any logical group, and perhaps they also will not ever be seen on the configuration page as well (if they do not have any user-facing parameters). Examples of this would be the tasks for splitting groups, and aggregating the final dashboard object. In these cases, those tasks can be left at the "top level" 

The order of all properties in the a group (and at the top level of the json schema as well) is recorded via the officially-supported `ui:schema` property defined by RJSF, as documented in https://rjsf-team.github.io/react-jsonschema-form/docs/json-schema/objects#specifying-property-order. 

Follow-on points:
- This feature is not reflected in the README, which as tracked in #217 was stale before this PR. In the interest of time and clarity of this PR, I will address that docs issue separately.
- The YAML forms do not reflect the groups because, as tracked in #47, they are currently generated through a separate code path. This is yet another motivation to fix #47,

cisaacstern:
Note also in the last two commits I've just tweaked the jsonschema generation slightly so that the keys are the task instance `id`s from the spec. This is so we can round-trip the formData returned from the frontend, and know what task `id`s the parameters should be used for.

For example, now `id: patrol_obs` here


https://github.com/wildlife-dynamics/ecoscope-workflows/blob/28f1ab5fd6673718abd5c224f9f40a1437c6c7ca/examples/compilation-specs/patrol_workflow.yaml#L15-L17

...gets used as the key in `"properties"` here (on line 26)...

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/28f1ab5fd6673718abd5c224f9f40a1437c6c7ca/examples/params/patrol_workflow_params_fillable.json#L22-L26

...so that the formData will look like this...

Screenshot 2024-09-13 at 3 21 21 PM


But by setting the `title`, we can still display a human-readable name:

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/28f1ab5fd6673718abd5c224f9f40a1437c6c7ca/examples/params/patrol_workflow_params_fillable.json#L80-L81



Screenshot 2024-09-13 at 3 20 46 PM

cisaacstern:
superseded by #277

PR #236, opened 2024-09-16
Add units to task results

Yun-Wu:
- **Add unit conversion**
- **update patrol related files**


atmorling (commented):
Looks good, have left a couple comments.

At first glance, it caught me by surprise that this converts the `float` input to a `str` output, so the only other thing I'd raise/question is whether the task should be called something like `format_unit_for_display`?
Just to make it clear that the intention is for adding to a widget and distinguish it from a unit conversion that might be done before a calculation.

ecoscope_workflows/tasks/transformation/_conversion.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@task
+def unit_convert(
+    number: Annotated[float, Field(description="The original value.")],
+    original_unit: Annotated[str, Field(description="The column name to explode.")],
+    new_unit: Annotated[str, Field(description="The column name to explode.")],  # todo
+    decimal_places: Annotated[int, Field(description="The column name to explode.")],
+) -> Annotated[str, Field(description="The min of the column")]:  # todo
+    import astropy.units as u
```
cisaacstern, line 18: My main reflection is wondering if we want to rely on something other than `astropy` for unit conversion; I'm not totally clear on the specifics, but `astropy` seems potentially a bit heavier of a dependency than we need here? While we do get it for "free" in the context of workflows that otherwise depend on `ecoscope` core, if others eventually want to reuse this task outside that context, it might be preferable to use a lighter-weight / less domain-specific dependency for that, perhaps https://pint.readthedocs.io/en/stable/, if that can be made to work (though I have not looked into the details yet).

That being said, if this is the easiest way to unblock unit conversion for now, happy to go this route and just open an Issue to track possibly re-implementing this in a different way down the line.

Thanks for working on this, Yun!

Yun-Wu: Thank you Charles! I did some research on the unit conversion packages and `pint` is also considered a heavy one. I eventually selected astropy mainly because ecoscope is already using it and I don't want to introduce yet another dependency. 

This is the summary by chatGPT. What do you think? 
截屏2024-09-19 下午5 43 22

ecoscope_workflows/tasks/transformation/_conversion.py
```diff
@@ -0,0 +1,22 @@
...
+from ecoscope_workflows.decorators import task
+
+logger = logging.getLogger(__name__)
+
+
+@task
+def unit_convert(
+    number: Annotated[float, Field(description="The original value.")],
+    original_unit: Annotated[str, Field(description="The column name to explode.")],
+    new_unit: Annotated[str, Field(description="The column name to explode.")],  # todo
```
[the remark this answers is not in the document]
Yun-Wu: yea, we can probably reused the literals the package also ready defined. Let me search for it.

ecoscope_workflows/tasks/transformation/_conversion.py
cisaacstern: Thanks for this additional context, Yun! I agree Astropy seems like the fastest path here at the moment, and we already have the dependency, so let's go with that. We can always implement a different unit-converter task using one of the others if there is a need for unit conversion somewhere that astropy is not already installed.

cisaacstern:
> At first glance, it caught me by surprise that this converts the float input to a str output, so the only other thing I'd raise/question is whether the task should be called something like format_unit_for_display?

A-ha just now understanding that this is the goal.

With that in mind, the other route I'd suggest we consider is a structured result? So rather than returning `{"value": "1 km"}` should we instead be considering `{"value": "1", "unit": "km"}` ?

This is actually already provided for in the server, here:

https://github.com/wildlife-dynamics/ecoscope-server/blob/c8ec38e7f94a087d1c05e2c6a9e58916199fc7ca/ecoscope_server/schemas/workflow_layout.py#L97-L100

And is the way @juanlescano-ng had done things prior to my starting.

This seems perhaps preferable? Among other things it would allow the frontend to be more opinionated about how to style the result (different font/size/weight for the value vs. the unit, perhaps).

ecoscope_workflows/tasks/transformation/_conversion.py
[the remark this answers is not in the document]
Yun-Wu: @atmorling There are actually tons of units supported by the package: https://eteq.github.io/astropy/units/index.html#module-astropy.units.si including length, time, distance, area, volume, velocity, acceleration, pressure, etc. If we explicitly define them, there may be too many options in the drop-down list. They have a pretty good parser to understand the units, including "m" "meter" "meters" "metre", etc. But I agree it's less error-prone with an enum. There are upsides and downsides to each way of doing it. What do you think?

PR #257, opened 2024-09-23
Pixi build directories for workflows (2)

cisaacstern:
Reopening #225 because for some reason when I changed the base of #225 from `main` to `cs/vipingo`, the GitHub Workflows stopped running.


cisaacstern:
Hmm this did not trigger the workflows 🤔 maybe it's because it can't run against `cs/vipingo` ?

cisaacstern:
Hmm maybe something to do with the `test-io` label 🤔  ... "CI" checks are running now but not "Examples". Strange.

PR #265, opened 2024-09-24
Add microservice + containerize apps

### LINKED ISSUE #246: Script runner microservice

cisaacstern:
closes #246


cisaacstern:
The big (and surprising) insight I had while working on this is that we now have the ability to generate a pydantic model for the workflow parameters. This solves the problems of making the signature of the workflow service being _way_ more expressive, and also let's us move `.validate` out of the decorator (which has always felt sort of awkward / improperly factored).

Next steps:

- [ ] Pull in the parts of #227 that fixed the keys in the json schema to be task instance ids rather than names
- [ ] Maybe just title the params model "Params"
- [ ] Integrate datamodel-codegen python api into compiler / artifacts
- [ ] In signatures of template `main` functions, use the model compiled by dm-codegen as the type hint for the `params`, and pass the params for each task by dotted attribute rather than dict key
- [ ] Remove .validate from templates and decorator
- [ ] Parse params in click app here

    https://github.com/wildlife-dynamics/ecoscope-workflows/blob/582e73647a61349f344dd633b19235f5929a226f/examples/mode-map/ecoscope-workflows-mode-map-workflow/ecoscope_workflows_mode_map_workflow/main.py#L38

cisaacstern:
Next steps:
- [ ] Move all artifacts to jinja to make it easy to add file headers to everything (test files, dockerfile, dockerignore, pixitoml, pyproject.toml, etc.)
- [ ] Add readme with graph png

cisaacstern:
Remaining work:
- [ ] Get docker tests to pass in CI (they work for me locally). Perhaps just a matter of waiting for container to warm up. Could sleep a few seconds before invoking it, or add a health check.
- [ ] Re-enable async tests
- [ ] finish dev docs
- [ ] Ticket out follow-on items from https://github.com/wildlife-dynamics/ecoscope-workflows/pull/265#issuecomment-2380300645

PR #283, opened 2024-10-04
Attach units to single_value results

Yun-Wu:
The major changes in this PR are:
- `_unit.py` attach units to single_value results
- `_widget_task.py` format single_value
- Specs in `patrols` and `subject_tracking` with `with_unit` tasks
- Fix workflow related tasks in `pixi.toml`


src/ecoscope-workflows-core/tests/tasks/test_widgets.py
```diff
@@ -66,7 +80,49 @@ def test_create_single_value_widget_single_view():
...
+def test_create_single_value_widget_single_view_decimal_places():
+    title = "A Great Value"
+    view = (("month", "=", "january"), ("year", "=", "2022"))
+    data = Quantity(value=123.45, unit=Unit.HOUR)
+
+    widget = create_single_value_widget_single_view(title, data, view, decimal_places=3)
+    assert widget == WidgetSingleView(
+        widget_type="stat",
+        title=title,
+        data="123.450 h",
```
cisaacstern, line 125: Very nicely expressive tests, thanks for always including such great representative examples in the unit tests 🙏

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/transformation/_unit.py
```diff
@@ -0,0 +1,52 @@
...
+        return Quantity(value=value)
+
+    if not new_unit:
+        return Quantity(value=value, unit=original_unit)
+
+    import astropy.units as u  # type: ignore[import-untyped]
+
+    original = value * u.Unit(original_unit.value)
+    new_quantity = original.to(u.Unit(new_unit.value))
+    return Quantity(value=new_quantity.value, unit=new_unit)
```
cisaacstern, line 52: Nice simple API 😃

cisaacstern (approved):
LGTM, @Yun-Wu!

Having reviewed this, I really feel how your suggestion #284 will be a very big developer experience boost.

I tried to do a quick for that in #285 and merged that here just to see if it would work. It did not, so I left #284 open. I _think_ #285 was implemented correctly per the GitHub docs, so my hope (as noted there) is that the generated files were not omitted here only because that cannot be changed after the PR has already been opened. 🤞 The next time we open a PR with generated files, #285 will "just work".

Anyway, looks great please merge when ready! 🙏

PR #287, opened 2024-10-05
Enforce PRs Linked with Issues

### LINKED ISSUE #288: Enforce PRs Linked with Issues

Yun-Wu:
Fix #288

### DIFF
--- a/.github/workflows/link_issue.yml
+++ b/.github/workflows/link_issue.yml
@@ -27,12 +27,18 @@ jobs:
             const { owner, repo } = context.repo;
 
             try {
-              await github.rest.issues.get({
+              const issue = await github.rest.issues.get({
                 owner,
                 repo,
                 issue_number: issueNumber
               });
-              console.log(`Linked issue #${issueNumber} found and verified.`);
+              const minLength = 30; // Minimum description length
+
+              if (!issue.data.body || issue.data.body.trim().length < minLength) {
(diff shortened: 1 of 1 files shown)


cisaacstern (approved):
This is so cool!

Do you think we should also require that the linked issue has a description in its opening comment?

I note that CI is passing here because of being linked to #288, but that issue doesn't have a description in its opening comment: 

Screenshot 2024-10-06 at 10 54 47 AM

I don't have a very strong opinion on this, just wondering if that extra step may provide even more clarity.

Yun-Wu:
> I note that CI is passing here because of being linked to #288, but that issue doesn't have a description in its opening comment:

Good question. I added this PR mainly for the PGAFF reporting purpose and we don't need descriptions there. I agree adding descriptions could make thing clearer but also want to minimize the extra overhead for developers. What do you think @walljcg and @atmorling ?

walljcg:
Thanks @Yun-Wu for this! Yes, I think we should force a description! I might regret that at some point but will help us all I think. Would be great if we can enforce this behaviour across all ecoscope repos including the ecoscope-pgaff reop.

atmorling:
I agree with @walljcg - I think we're all going to hate this at some point or another but being forced into actually documenting what I'm doing seems like the right idea :sweat_smile: 

Thanks @Yun-Wu for making this happen!

PR #294, opened 2024-10-06
Verify build tag for releases

### LINKED ISSUE #293: Fix publish workflow
Prioritizing because this is a blocker for https://github.com/ecoscope-platform-workflows-releases/patrols/pull/1

Current issues observed from latest attempt to publish following #292:
- [ ] `rattler-build` not available in `publish-to-prefix` job. this can be fixed by either manual install in workflow, or (probably better) implementing this as a pixi task
- [ ] tagged release version is not propagated through to `publish/recipes/*.yaml` ... or at least, there is no check for this. this can be the developer's responsibility but we can reject builds that don't have this correct by failing out with verification. we have all the dependencies we need (namely, i think `yq`) to check the recipe version against the github actions trigger ref/tag.

cisaacstern:
closes #293


cisaacstern:
Really only (easily) possible to test this by merging it to main and trying to release, so going to do that now.

PR #303, opened 2024-10-08
App response model

### LINKED ISSUE #304: Stricter typing for app response model
In working on https://github.com/wildlife-dynamics/compose/pull/16 I found that it will be very useful if we can be more expressive through typing as to the data returned from the app.

Implemented in #303

cisaacstern:
In working on https://github.com/wildlife-dynamics/compose/pull/16, I found that it would be very helpful to actually leverage typing to convey how the dashboard json model fits into the app response model. This PR does that.

closes #304


cisaacstern:
Note, in being more opinionated about the response model, I've ended up needing to delete the mode-map and mode-mapvalues examples, as they did not conform to that model. We can bring them back in some other form (tutorial docs?) if needed. I do kind of like reducing the number of examples in this repo, though.

PR #312, opened 2024-10-11
Improve patrol and subject tracking workflows

### LINKED ISSUE #232: Map Task: title, base layer, etc
1. support multiple base layers
2. support opacity
3. support camera
4. north arrow sizing does not work??? Need to test again

### LINKED ISSUE #280: Subject Tracking Workflow - Speed Map
Add a good old speed map to the subject tracking workflow
- should use the green (slow) to red (fast) colour ramp
- max speed should be from the overall trajectory (i.e., relative to population rather than individual)

Yun-Wu:
- Add speedmap to subject tracking workflow. Closes #280 
- Apply color map to time density, patrol events, etc.

- Allow default widgets for Ecomap. Closes #232 
- Support a list of tile layers with opacity settings. Closes #232 

- Fix time_density calculation bug
- Fix legend issue where color cannot be correctly applied
- Fix apply_classification data model issue

 
- Update patrol workflows to use the latest generated data


src/ecoscope-workflows-core/ecoscope_workflows_core/testing.py
```diff
@@ -85,7 +85,7 @@ def run_cli_test_case(
                 "backend": "localhost",
                 "storage": "localhost",
                 "log_level": "INFO",
-                "data_limit": 16,
+                "data_limit": 256,
```
cisaacstern, line 88: I will do a full review tomorrow but because I know you'd asked about this specifically, this seems fine to me! Good catch!

src/ecoscope-workflows-ext-ecoscope/ecoscope_workflows_ext_ecoscope/tasks/results/_ecomap.py
```diff
@@ -117,9 +121,10 @@ def draw_ecomap(
         LayerDefinition | list[LayerDefinition],
         Field(description="A list of map layers to add to the map.", exclude=True),
     ],
-    tile_layer: Annotated[
-        str, Field(description="A named tile layer, ie OpenStreetMap.")
-    ] = "",
+    tile_layers: Annotated[
+        list[TileLayer],
+        Field(description="A list of named tile layer with opacity, ie OpenStreetMap."),
+    ] = [],
```
cisaacstern, line 127: Wow I was about to say we should use default_factory here, which in any non-pydantic situation would be preferable/required, but [TIL that is not necessary with pydantic](https://docs.pydantic.dev/latest/concepts/models/#fields-with-non-hashable-default-values) 😄

cisaacstern (approved):
Wow @Yun-Wu thank you so much for this incredibly thorough PR. Apologies for the delay. Typically I know you prefer to merge yourself, but I do need this tonight to unblock https://github.com/wildlife-dynamics/compose/pull/53 so going to merge! Hope that's alright 🙏 and TYSM again!!

PR #356, opened 2024-10-30
Workflow details

### LINKED ISSUE #355: Workflow Details task for all three example workflows
We need a task that renders [the Workflow Details model from Figma](https://www.figma.com/proto/0W4QLZWryZdOqgYHpwrr9H/Ecoscope?node-id=3824-16512&starting-point-node-id=3825%3A18781) at the top of each configuration form. This can then be used as a setter for the dashboard metadata fields (rather than those being set directly).

This is a blocker for https://github.com/wildlife-dynamics/ecoscope-web/pull/128, because that PR needs to know these details (i.e. name, description, image_url) to make the POST request for creating a new workflow. Note that to fulfill the frontend requirement, the task instance id for this task must always be the same (so that the frontend knows which part of the formdata to extract these values from).

Todo:

- [ ] `set_workflow_details` task
- [ ] include this task in all examples (with a reserved/the same task id name; e.g. "workflow_details")
- [ ] cut a new prefix release that includes this task
- [ ] recompile and re-deploy all workflows releases
- [ ] circle back to frontend PR and close that out

atmorling:
closes #355


atmorling:
Updated RJSF looks like this: 
```json

```

PR #368, opened 2024-11-01
Plot title kwargs

### LINKED ISSUE #366: Handle title kwargs in ecoplot
@Temutev noticed that the ecoplot tasks don't support setting a title

We should surface the title kwargs from `go.layout` here

atmorling:
WIP
Closes #366


cisaacstern (changes requested):
@atmorling apologies for the delayed review.

I am marking as "Request changes" to make sure we don't accidentally deploy this before discussing the presence of these additional args in the RJSF.

Is that deliberate? It would seem to me the title in particular would result perhaps in another "duplicate title" scenario like we saw with ecomap? (Title on the iframe container + title on the html itself.)

As for the other two args (title_y and title_x), are those axis titles, and if so do we want to be exposing them as "basic config" in the RJSF?

cisaacstern (approved):
Thanks for the clarification!

PR #378, opened 2024-11-05
Map layer tasks

atmorling:
Adds individual tasks for each point/polygon/polyline layer type


cisaacstern (approved):
A few docstrings nits which I will merge now. Otherwise LGTM, thanks Alex!

PR #384, opened 2024-11-06
More AdvancedField rjsf enhancements

### LINKED ISSUE #385: Merge latest rjsf enhancements
I'll get ci passing on the latest rjsf enhancements and merge those.

cisaacstern:
closes #385


src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/transformation/_indexing.py
```diff
@@ -16,8 +16,8 @@ def add_temporal_index(
         str, Field(description="The name of the column containing time data.")
     ],
     groupers: Annotated[
-        list[Grouper | TemporalGrouper],
-        Field(description="A list of groupers which may contain TemporalGroupers."),
+        list[TemporalGrouper],
+        Field(description="A list of TemporalGroupers."),
```
cisaacstern, line 20: I agree that the typing you propose here feels more accurate to the spirit of the task.

In the context of the workflows, however, we currently have all of the groupers in a single list, as emitted from `set_groupers`, so here we need to be able that heterogenous list of groupers here in the context of passing that return value in the context of the DAG.

Perhaps there is a more expressive way to type or document this (and as I write this I realize it should also be `exclude=True`, because it would never be set in RJSF, just in the spec.yaml).

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/transformation/_indexing.py
```diff
@@ -39,12 +39,12 @@ def add_temporal_index(
...
 
+    if cast_to_datetime:
+        df[time_col] = pd.to_datetime(df[time_col], format=format)
+
     temporal_groupers = [g for g in groupers if isinstance(g, TemporalGrouper)]
 
     for tg in temporal_groupers:
-        if cast_to_datetime:
-            df[time_col] = pd.to_datetime(df[time_col], format=format)
-
```
cisaacstern, line 47: Yeah this makes more sense! Good catch.

cisaacstern (commented):
Looking good! I will push a few changes to get tests to pass and then merge.

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/transformation/_indexing.py
cisaacstern: I 💯 agree that we should only cast_to_datetime once. Just made a minor tweak to this so that we don't even bother doing that if `temporal_groupers` is empty.

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/transformation/_indexing.py
cisaacstern: I restored the original typing and added a much longer description to explain the motivation there.

PR #386, opened 2024-11-06
More patrols rjsf, and add `set_connection` task

### LINKED ISSUE #387: Add `set_connection` task to reuse same named connection in multiple places
In patrols, for example, we want to reuse the same connection in multiple i/o tasks, so let's set that once.

Only thing that makes this a little tricky is we need to maintain compatibility with parsing enums on the server side.

cisaacstern:
closes #387
towards https://github.com/wildlife-dynamics/compose/issues/131


cisaacstern:
looks like this is working mostly as expected now.

just need to look into:

- server parsing of named connections enum mentioned in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/387#issue-2639141522
- maybe also run this without mock_io as a sanity check (though frankly can't imagine why it won't work)

cisaacstern:
> server parsing of named connections enum mentioned in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/387#issue-2639141522

Done (though admittedly not tested) in [7dd93ef](https://github.com/wildlife-dynamics/ecoscope-workflows/pull/386/commits/7dd93ef895f875dde6f6c14c722de91b66af03b3)

> maybe also run this without mock_io as a sanity check (though frankly can't imagine why it won't work)

Perhaps I will regret this, but I truly cannot envision any reason this will fail, and the effort to test this is in a way greater than just deploying it and seeing if it works.

I wouldn't consider releasing an untested change like this if we had active users, but since we are not in that place right now, I think it's okay to just merge this, and then circle back if we have any issue (which I will be amazed if we do).

PR #395, opened 2024-11-08
Rename set_connection `name` param to globally unique, reserved name

### LINKED ISSUE #396: Use globally unique name for `name` param of set_connections task
As described in https://github.com/wildlife-dynamics/compose/issues/157, we need a globally unique name for the set_connections `name` param that can be used to identify it in the server for parsing the named connections at workflow trigger time.

cisaacstern:
closes #396


cisaacstern:
Some thoughts here:
- Paired with https://github.com/wildlife-dynamics/ecoscope-server/pull/169
- I've already tagged 0.0.17 so we can tag right after this
- The reason so many files are changed is because, much to my surprise, I realized that every user-facing parameter name actually does appear in the Spec.model_dump_json() output (and therefore affects the sha256 hash of the workflow)

PR #425, opened 2024-11-13
Use subject name as label in NSD plot 

### LINKED ISSUE #394: Subject tracking NSD chart should use subject name (not id) as lablel
@walljcg realized that this labelling is not that readable

Image

Jake thinks this should be fixed at the `draw_ecoplot` labeling level (e.g. it's maybe not specific to NSD)

atmorling:
Closes #394


atmorling:
As you can see in the spec change it looks like I'm referring to two different columns,
in working through the fix for this I had to trace through the workflow to figure out what happens to the column name:

`get_subjectgroup_observations` returns with `extra__subject__name`
`process_relocations` turns `extra__subject__name` into `name`
`relocations_to_trajectory` re-adds extra so from `name` to `extra__name`

To me this highlights the need for a rethink of our column naming in ecoscope core and something to look at with a refactor of the base Relocs and Traj objects, if not sooner...

PR #427, opened 2024-11-13
In `get_patrol_events`, truncate to time range

### LINKED ISSUE #426: Patrol events & observations may be misaligned in time dimension
Because events associated with a patrol may fall outside the time range used to query the observations, there can be a misalignment in the time dimension which is causing failures in patrols workflow.

cisaacstern:
closes #426


cisaacstern:
As part of this PR I have also defaulted `status` to `["done"]` so the default for this field matches that of `get_patrol_observations`. This can be overridden with an inline default in the spec if necessary / desired.

PR #449, opened 2024-11-18
Bump 0.0.23 and ecoscope

### LINKED ISSUE #447: Release 0.0.23
In order to push through the latest ecoscope core changes

atmorling:
Closes #447


examples/subject-tracking/ecoscope-workflows-subject-tracking-workflow/pixi.lock
```diff
@@ -276,11 +276,11 @@ environments:
...
       - conda: https://conda.anaconda.org/conda-forge/linux-64/openssl-3.4.0-hb9d3cd8_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/orc-2.0.2-h669347b_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/packaging-24.2-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/packaging-24.2-pyhff2d567_1.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/palettable-3.3.3-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/pandas-2.2.3-py312hf9745cd_1.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/pandera-0.20.4-hd8ed1ab_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/pandera-base-0.20.4-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pandera-0.21.0-hd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pandera-base-0.21.0-pyhd8ed1ab_0.conda
```
cisaacstern, line 283: Possibly unlikely but not impossible this pandera version bump is changing the validation behavior in such a way as to cause the failure we're seeing in CI

cisaacstern: Linking for ease of navigation https://github.com/wildlife-dynamics/ecoscope-workflows/actions/runs/11900163896/job/33160504564?pr=449#step:6:220

cisaacstern:
@atmorling, I looked into the pandera thing. It looks like [this was the PR](https://github.com/unionai-oss/pandera/pull/1821) included in the latest pandera release which is causing the issues here. The problem seems to be that lookups for the "backend type" in pandera were optimized so as not to require `isinstance` checks and instead now use lookups based on the string names of base classes for the object to be validated.

That's all good, but it seems the implementation released in the lastest pandera doesn't account for more-than-depth-1 inheritance. I have a fix here that perhaps makes it more clear what I mean by that: https://github.com/unionai-oss/pandera/pull/1857/files

Running from this diff locally seems to resolve the problem we are seeing (at the unit-testable level).

So our fastest option to move forward here probably is to pin to an upper bound of the prior release of pandera, and then I will try to move forward the fix in pandera upstream so we can release that pin.

cisaacstern:
> So our fastest option to move forward here probably is to pin to an upper bound of the prior release of pandera, and then I will try to move forward the fix in pandera upstream so we can release that pin.

The alternative is we could release an "internal" version of pandera to our prefix channel using my diff linked above. This keeps us ahead of any other changes in pandera are on their way (but requires a smidge more work).

atmorling:
Have pushed an upper bound on Pandera

PR #452, opened 2024-11-20
Auto-recompile workflows with the help of `ecoscope-elebot`

### LINKED ISSUE #432: Github Workflow to auto-recompile examples when certain files change
Developers should need to recompile workflows manually.

Let's do that for them (ourselves) with a GitHub workflow that functions something like precommit.ci.

When certain files change, e.g. the compiler / artifacts logic, examples/spec.yaml or examples/test-cases.yaml, etc., then we can have a Github Workflow re-compile and push those changes as a commit on the back of the developer's changes. This could be disabled by a PR label flag or similar.

### LINKED ISSUE #439: Make recompiles even faster with --no-install
Splitting this off from https://github.com/wildlife-dynamics/ecoscope-workflows/issues/431 because its a bigger effort than we have bandwidth for now. See #431 for discussion.

### LINKED ISSUE #450: Bumping ecoscope requires `lock-ext-ecoscope`
Bumping ecoscope requires `lock-ext-ecoscope` in order to bump `src/ecoscope-workflows-ext-ecoscope/pixi.lock`
This is seemingly required to get `recompile-all` to work - otherwise we hit:
```bash
✨ Pixi task (compile-events): ./dev/recompile.sh events --install
  × lock-file not up-to-date with the project

recompiling 'examples/events/spec.yaml' with flags '--clobber --install'
  × lock-file not up-to-date with the project
```

Steps to reproduce:
- Bump ecoscope in `src/ecoscope-workflows-ext-ecoscope/pyproject.toml`
- Attempt to recompile/compile

cisaacstern:
closes #450
closes #439
closes #432


cisaacstern:
For why events inline defaults are being changed here, see discussion in https://github.com/wildlife-dynamics/ecoscope-workflows/issues/459#issuecomment-2489827408. This appears to be a "rectification" of the compiled artifacts against what the committed spec.yaml defines. If that is true, that is a very good endorsement of automating this.

cisaacstern:
A few notes for context and perhaps to lift into future documentation:

- Dropping the workflow package pyproject.toml and associated editable pypi-dependencies is to address https://github.com/prefix-dev/pixi/issues/817. This seems pretty lightweight of a change in the scheme of things actually, and actually makes our artifacts simpler (to a degree) which is good.
- Recompile may be stymied by lockfile conflicts. In this case, the developer can manually checkout the lockfile from `main`, commit that, and then ecoscope-elebot will take it from there.

publish/build.sh
```diff
@@ -26,6 +25,7 @@ fi
 echo "You selected: $1"
 echo "Building recipes: ${RECIPES[@]}"
 
+rm -rf /tmp/ecoscope-workflows/release/artifacts
```
atmorling, line 28: :heart:

atmorling (approved):
Looks great! TYSM 
Just so I understand the flow of events correctly, this doesn't care what state the pushes are in, I could recompile events only, or all workflows, or none, and the bot is going to blanket recompile over the top regardless?

cisaacstern:
> Just so I understand the flow of events correctly, this doesn't care what state the pushes are in, I could recompile events only, or all workflows, or none, and the bot is going to blanket recompile over the top regardless?

@atmorling, correct. And in fact, it's now recommended you _not_ recompile anything, as the bot will do that for you, and you can always pull the changes from remote after that's done. But if you _do_ recompile something, that will be over-written by the bot as you describe.

cisaacstern:
@atmorling and others, please also note this comment from above:

> Recompile may be stymied by lockfile conflicts. In this case, the developer can manually checkout the lockfile from main [or the base branch, whatever that may be], commit that, and then ecoscope-elebot will take it from there.

Here, the "take it from there" is for the same reason you describe above, Alex. The bot will overwrite all compiled artifacts on every commit, no matter what.

cisaacstern:
>  In this case, the developer can manually checkout the lockfile from main [or the base branch, whatever that may be], commit that, and then ecoscope-elebot will take it from there.

For example, in the case of lockfile conflicts with main

```console
$ git checkout origin/main -- **/**/pixi.lock
$ git commit
```

atmorling:
Are we referring to local recompiles here or bot recompiles, or both?

cisaacstern:
> Are we referring to local recompiles here or bot recompiles, or both?

A-ha good question. Bot recompiles is what I meant. It's been ages since I've locally recompiled so not sure if that is affected or not. (By ages I mean hours, since this has been working.)

PR #457, opened 2024-11-20
Hide sort values in SubjectTracking RJSF

### LINKED ISSUE #453: sort_values should be hidden from rjsf in example workflows
This can be fixed with an inline default, but we should also check the task params to flag anything that should be `AdvancedField`

atmorling:
Closes #453


atmorling:
I've left `ascending` as a non-advanced field but have inlined defaulted it.
`na_position` is definitely an advanced field so have tagged as such

Issue #151, opened 2024-08-06
Dashboard filters: enumNames is deprecated in RJSF


cisaacstern:
Per @juanlescano-ng this is deprecated:

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/e596df3d66159e10f0f059a69b422c957cdf3a7b/ecoscope_workflows/jsonschema.py#L40

whoisjuandev:
> enumNames is a non-standard JSON Schema field that was deprecated in version 5

Docs to support this deprecated property: https://rjsf-team.github.io/react-jsonschema-form/docs/migration-guides/v5.x%20upgrade%20guide/#non-standard-enumnames-property

PR #461, opened 2024-11-20
Temporal grouper index rjsf

### LINKED ISSUE #446: Restrict TemporalGrouper RJSF to single enum of self-describing options
We should limit the input fields for TemporalGroupers to a single Enum of describing directives (with readable display names).

atmorling:
closes #446


src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
```diff
@@ -67,6 +75,13 @@
 ]
 
 
+TemporalIndexSelector = tuple([mapping["value"] for mapping in strftime_directives])
+TemporalIndexJsonSchemaExtra = {
+    "enum": [mapping["value"] for mapping in strftime_directives],
+    "enumNames": [mapping["display"] for mapping in strftime_directives],
+}
```
cisaacstern, line 82: enum/enunNames is apparently deprecated in RJSF, see https://github.com/wildlife-dynamics/ecoscope-workflows/issues/151

Which, if memory serves, is why we went with the [oneOf implementation like this](https://github.com/wildlife-dynamics/ecoscope-workflows/blob/2989e5b247ad1a6ca58d74a6477d8a19026c163a/src/ecoscope-workflows-core/ecoscope_workflows_core/jsonschema.py#L46-L55)

For consistency, I would suggest we leverage that existing definition. Combining this with the idea above of iterating on keys/values, I would suggest we do the following (also, with a snake case variable name since this is an instance of oneOf):

```suggestion
from ecoscope_workflows_core.jsonschema import oneOf

temporal_index_json_schema_extra = oneOf(
    const=[k for k in strftime_directives.keys()],
    title=[v for v in strftime_directives.values()],
).model_dump()
```

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
```diff
@@ -82,8 +97,10 @@ class Grouper:
 # subclasses cannot set required fields if their parent class has default fields defined.
 @dataclass(frozen=True)
 class TemporalGrouper:
-    index_name: IndexName
-    directive: Directive
+    index_name: Annotated[
+        Literal[TemporalIndexSelector],
+        Field(json_schema_extra=TemporalIndexJsonSchemaExtra),
```
cisaacstern, line 102:
```suggestion
        Field(json_schema_extra=temporal_index_json_schema_extra),
```

cisaacstern (commented):
Looks like Field(json_schema_extra=...) is a good way into this.

I made some implementation/stylistic suggests below.

Before we merge this we should also figure out what specific subset of these directives we want to support, and for those, use the better text (plus example) from https://strftime.org/.

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
cisaacstern: I'm not certain if the above code runs, but I imagine if it doesn't, it's quite close to being able to.

cisaacstern: We may need to use the key `"oneOf"` as we do for the dashboard filters here

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/2989e5b247ad1a6ca58d74a6477d8a19026c163a/src/ecoscope-workflows-core/ecoscope_workflows_core/jsonschema.py#L69

so maybe something like 

```python
temporal_index_json_schema_extra = {
    "oneOf": oneOf(
        const=[k for k in strftime_directives.keys()],
        title=[v for v in strftime_directives.values()],
    ).model_dump()
}
```

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
```diff
@@ -14,31 +15,36 @@
...
-    "%x": "Locale's appropriate date representation.",
-    "%X": "Locale's appropriate time representation.",
-    "%y": "Year without century as a decimal number [00,99].",
-    "%Y": "Year with century as a decimal number.",
-    "%z": "Time zone offset indicating a positive or negative time difference from UTC/GMT of the form +HHMM or -HHMM.",
-    "%%": "A literal '%' character.",
-}
+strftime_directives = [
+    {"%a": "Locale's abbreviated weekday name."},
+    {"%A": "Locale's full weekday name."},
```
cisaacstern, line 20: you probably realized this but the ci is failing bc this is list of dicts rather than just a dict.

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
```diff
@@ -67,6 +67,13 @@
 ]
 
 
+TemporalIndexSelector = tuple(value for value in strftime_directives.values())
+temporal_index_json_schema_extra = oneOf(
+    const=[k for k in strftime_directives.keys()],
+    title=[v for v in strftime_directives.values()],
+).model_dump()
```
cisaacstern, line 74:
```suggestion
temporal_index_json_schema_extra = {
    "oneOf": [
         oneOf(const=k, title=v).model_dump()
         for k, v in strftime_directives
    ]
}
```

oops i lead you astray with this recommendation, it should be something like this

atmorling: no stress! on it :smile:

atmorling: Local recompile definitely seems quicker for iterating on things like this

cisaacstern: fair!

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
```diff
@@ -89,7 +94,8 @@ class Grouper:
 @dataclass(frozen=True)
 class TemporalGrouper:
     index_name: Annotated[
-        Literal[TemporalIndexSelector],
+        str,
+        AfterValidator(validate_temporal_index),
```
atmorling, line 98: This is necessary because the `enum` generated by `Literal[TemporalIndexSelector]` creates conflicts with `oneOf`

atmorling (commented):
This works, but introduce some potentially confusing behaviour in the RJSF such that if a user does the following:

- Selects Temporal Grouper from the Grouper drop-down
- Selects a Temporal Index 
- Switches the Grouper drop-down to a Value Grouper
The "value" of the selected Temporal Index (ie "%B") will be visible in the free text field for Value Grouper.

This is because `index_name` is shared between the two Grouper options, AND because the `IndexName` type ([here](https://github.com/wildlife-dynamics/ecoscope-workflows/blob/f9d5086e5e04d6555f68b49395e743fef6dd5f6f/src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py#L71-L72)) is just any `str`

I don't think this is a dealbreaker but it definitely highlights the need for a similar "shift letf" lift on Value Groupers

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
cisaacstern: Can you give an example?

cisaacstern:
> Then the "value" of the selected Temporal Index (ie "%B") will be visible in the free text field for Value Grouper.
> This is because index_name is shared between the two Grouper options

I think if both elements of the union type have the same args, pydantic may struggle to discriminate accurately between them on parse? Perhaps the best solution is just to choose different names for these args?

atmorling:
> > Then the "value" of the selected Temporal Index (ie "%B") will be visible in the free text field for Value Grouper.
> > This is because index_name is shared between the two Grouper options
> 
> I think if both elements of the union type have the same args, pydantic may struggle to discriminate accurately between them on parse? Perhaps the best solution is just to choose different names for these args?

:eyes: Nice intuition! That's exactly why the last CI run failed.

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
atmorling: So using this 
```python
    index_name: Annotated[
        Literal[TemporalIndexSelector],
        Field(json_schema_extra=temporal_index_json_schema_extra),
    ]
```
Generates [this](https://gist.github.com/atmorling/4106952062bb1288fb8c6d1f73373a8a) with `enum` and `oneOf`

Playing around with it some more just now, and knowing what I didn't realise at the time, which is that Pydantic struggles to discriminate, I think this is also true for the form renderer and is what is happening here. I'm gonna fix the arg name and then revert this as a test.

atmorling: Realised I could just test that in RJSF.... that does seem to hold true

atmorling: In the above example, the Literal is from the display values. This is obviously fine in RJSF land but means that if one is ever defining TemporalGroupers in Python, one would have to provide "Month (September)" rather than "%B"

Unfortunately building the Literal from the directives causes issues with RJSF because the list of values in `enum` is now different to the list of `title`'s in `oneOf` ie:
```json
"index_name": {
  "enum": [
    "%a",
    "$A",
    "%b"
  ],
  "oneOf": [
    {
      "const": "%a",
      "title": "Locale's abbreviated weekday name."
    },
    {
      "const": "%A",
      "title": "Locale's full weekday name."
    },
    {
      "const": "%b",
      "title": "Locale's abbreviated month name."
    },
  ],
  "title": "Index Name",
  "type": "string"
}

```

cisaacstern: This feels to me like we shouldn't be "doubling up" on this definition... like maybe we want to use WithJsonSchema so we don't need to deal with the enum at all?

atmorling: I think you're right, there's also not that much to the generated schema so "overriding" it doesn't seem too brittle

src/ecoscope-workflows-core/ecoscope_workflows_core/jsonschema.py
```diff
@@ -54,6 +54,9 @@ class oneOf(BaseModel):
     const: Any
     title: str
 
+    def __hash__(self) -> int:
+        return self.const.__hash__()
```
cisaacstern, line 58: Maybe something like

```suggestion
    def __hash__(self) -> int:
        return hash(("oneOf", self.const, self.title))
```

?

atmorling: Sure!

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
```diff
@@ -16,18 +16,16 @@
...
-    "%A": "Day of the week (Sunday)",
-    "%H": "Hour (24-hour clock) as number",
-    "%M": "Minute",
-    "%S": "Second",
+    "%Y": "Year (Example: 2024)",
+    "%B": "Month (Example: September)",
+    "%j": "Day of the year as a number (Example: 365)",
+    "%d": "Day of the month as a number (Example: 31)",
+    "%A": "Day of the week (Example: Sunday)",
+    "%H": "Hour (24-hour clock) as number (Example: 22)",
```
cisaacstern, line 24:
```suggestion
    "%Y": "Year (example: 2024)",
    "%B": "Month (example: September)",
    "%j": "Day of the year as a number (example: 365)",
    "%d": "Day of the month as a number (example: 31)",
    "%A": "Day of the week (example: Sunday)",
    "%H": "Hour (24-hour clock) as number (example: 22)",
```

cisaacstern:
Let's make sure we are still getting a good display name on the frontend

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/c0195cc246f3a7fa4d0a1015dd44fa4dee5b469b/src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/results/_dashboard.py#L154

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
```diff
@@ -82,11 +50,16 @@ class Grouper:
...
 class TemporalGrouper:
-    index_name: IndexName
-    directive: Directive
+    temporal_index: Annotated[
+        TemporalIndexKeys, WithJsonSchema(temporal_index_json_schema)
+    ]
     display_name: Annotated[
         str | SkipJsonSchema[None],
         AdvancedField(default=None, description="A human-readable name for the index."),
     ] = None
```
cisaacstern, line 59: Let's make this a property

```python
@property
def display_name(self):
    return strftime_directives[self.index_name].split("(")[0].strip()

```

cisaacstern: Per Alex suggestion, let's make the directives frozen dataclasses! 🎉

examples/events/test-cases.yaml
```diff
@@ -13,9 +13,10 @@ test1:
     # Parameters for 'Set Groupers' using task `set_groupers`.
     groupers:
       groupers: # (list[ecoscope_workflows.tasks.groupby._groupby.Grouper], FieldInfo(annotation=NoneType, required=True, description='            Index(es) and/or column(s) to group by, along with\n            optional display names and help text.\n            '))
-        - index_name: "month"
-          display_name: "Month"
-          directive: "%B"
+        - temporal_index:
+            directive: "%B"
+            selector_title: "Month (Example: September)"
+            display_name: "Month"
```
cisaacstern, line 19: @atmorling I would expect this to reflect 1:1 what the RJSF input needs to be.

```suggestion
        - temporal_index:
            selector_title: "Month (example: September)"
```
?

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
```diff
@@ -1,70 +1,76 @@
...
+    directive: Literal["%A"] = "%A"
+    selector_title: str = "Day of the week (Example: Sunday)"
+    display_name: str = "Day of the week"
+
+
+@dataclass(frozen=True)
+class Hour:
+    directive: Literal["%H"] = "%H"
+    selector_title: str = "Hour (24-hour clock) as number (Example: 22)"
+    display_name: str = "Hour"
```
cisaacstern, line 55: Lowercase `e` for example would be my preference, sry to be picky 🫣

examples/patrols/test-cases.yaml
```diff
@@ -13,10 +13,10 @@ test1:
...
     groupers:
       groupers: # (list[ecoscope_workflows.tasks.groupby._groupby.Grouper], FieldInfo(annotation=NoneType, required=True, description='            Index(es) and/or column(s) to group by, along with\n            optional display names and help text.\n            '))
-        - index_name: "month"
-          display_name: "Month"
-          help_text: "The month in which the patrol began or the event was updated."
-          directive: "%B"
+        - temporal_index:
+            directive: "%B"
+            selector_title: "Month (Example: September)"
+            display_name: "Month"
```
cisaacstern, line 19:
```suggestion
        - temporal_index:
            selector_title: "Month (Example: September)"
```
?

examples/subject-tracking/test-cases.yaml
```diff
@@ -13,9 +13,10 @@ test1:
     # Parameters for 'Set Groupers' using task `set_groupers`.
     groupers:
       groupers: # (list[ecoscope_workflows.tasks.groupby._groupby.Grouper], FieldInfo(annotation=NoneType, required=True, description='            Index(es) and/or column(s) to group by, along with\n            optional display names and help text.\n            '))
-        - index_name: "month"
-          display_name: "Month"
-          directive: "%B"
+        - temporal_index:
+            directive: "%B"
+            selector_title: "Month (Example: September)"
+            display_name: "Month"
```
cisaacstern, line 19:
```suggestion
        - temporal_index:
            selector_title: "Month (Example: September)"
```

?

src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py
atmorling: Skill issue on my part. I actually fixed this and then clearly blew it away

cisaacstern (commented):
@atmorling the gist of the test-cases suggestions I've made is that, if that yaml doesn't reflect 1:1 what the RJSF inputs need to be, then we're doing something wrong. That's kinda the idea behind test-cases.yaml: it shows us an example of what we expect the user to need to input.

atmorling:
> @atmorling the gist of the test-cases suggestions I've made is that, if that test-cases.yaml doesn't reflect 1:1 what the RJSF inputs need to be, then we're doing something wrong. That's kinda the idea behind test-cases.yaml: it shows us an example of what we expect the user to need to input.

Fair - was more basing it off what the returned form data would look like, which is distinct from the user input

cisaacstern:
> Fair - was more basing it off what the returned form data would look like, which is distinct from the user input

Will this even be what the formadata looks like, though? I think you mean the _parsed_ result of the `FormData` object, which is different from the returned form data.

atmorling:
Some of the mypy failures on teh latest are interesting: https://github.com/wildlife-dynamics/ecoscope-workflows/actions/runs/11979488364/job/33401775360?pr=461

The `_dashboard` failures should have reared it's head as soon as we changed `index_name` to `temporal_index` but has only surfaced now..

cisaacstern:
> The _dashboard failures should have reared it's head as soon as we changed index_name to temporal_index but has only surfaced now..

At one point I think the TemporalGrouper had `index_name` as a `@property` which seems to be gone now, and is why these errors are occuring?

atmorling:
Yep :facep

examples/events/test-cases.yaml
```diff
@@ -13,9 +13,8 @@ test1:
     # Parameters for 'Set Groupers' using task `set_groupers`.
     groupers:
       groupers: # (list[ecoscope_workflows.tasks.groupby._groupby.Grouper], FieldInfo(annotation=NoneType, required=True, description='            Index(es) and/or column(s) to group by, along with\n            optional display names and help text.\n            '))
-        - index_name: "month"
-          display_name: "Month"
-          directive: "%B"
+        - temporal_index:
+            selector_title: "Month (Example: September)"
```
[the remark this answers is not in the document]
cisaacstern: I am on mobile at the moment so hard to make a more specific suggestion. I can take a closer look when I'm back at the desk in a bit, if useful.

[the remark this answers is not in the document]
atmorling: The union type might be the problem, I've tried a few different cuts of yaml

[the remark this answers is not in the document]
atmorling: Looks like a mismatch between the Param model and the task input, 

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/9a852ed820437d43275c880b9c7e49af997fe8bd/examples/events/ecoscope-workflows-events-workflow/ecoscope_workflows_events_workflow/params.py#L70-L71

Params expects a string but our task is a Union Type of all the frozen dataclasses

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/9a852ed820437d43275c880b9c7e49af997fe8bd/src/ecoscope-workflows-core/ecoscope_workflows_core/indexes.py#L91-L95

davidovichmarcos (approved):

examples/events/test-cases.yaml
```diff
@@ -14,7 +14,9 @@ test1:
     groupers:
       groupers: # (list[ecoscope_workflows.tasks.groupby._groupby.Grouper], FieldInfo(annotation=NoneType, required=True, description='            Index(es) and/or column(s) to group by, along with\n            optional display names and help text.\n            '))
         - temporal_index:
-            selector_title: "Month (Example: September)"
+            directive: "%B"
+            selector_title:  "Year (example: 2024)"
+            display_name: "Year"
```
cisaacstern: A few comments:

- I think validation on the TemporalGrouper should not allow this combination? (`%B` and `Year...` are not compatible?)
- With the change to `type: "object"` below, would this also work?

```diff
-           directive: "%B"
            selector_title:  "Year (example: 2024)"
-           display_name: "Year"
```

I note that the variant of this above had capital `Example`, which may have been part of the problem.

cisaacstern (commented):
So close!

Two comments below, one re: validation, and the other is because the Params should reflect what the frontend sends.

cisaacstern (changes requested):
Tests are now passing but rjsf is not looking like it's rendering correctly in the playground, going to take a look at that now. Adding this "request changes" review to flag that this needs a further look before merge, despite passing tests.

cisaacstern (approved):
RJSF now works with latest changes.

PR #467, opened 2024-11-21
On release, automatically push changes to workflows releases

### LINKED ISSUE #465: On release, push changes to releases directories, and auto-recompile there
We can create a workflow that uses [the `on: release` trigger](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release) to watch for successful completions of [our publish workflow](https://github.com/wildlife-dynamics/ecoscope-workflows/blob/main/.github/workflows/publish.yml), and then:

1. Opens prs resetting the `ecoscope-platform-workflows-releases/*` repos to the current latest release tag. This entails:
    - Bumping top level pixi.toml to the lastest release
    - Bumping spec.yaml requirements sections to the latest release
    - Syncing any other changes from spec.yaml + test-cases.yaml
2. Those repos can have a version of the auto-recompile logic implemented in them, and ask @ecoscope-elebot to recompile and test them after these changes are opened as a PR. Tests can be made to run only for changes to the compiled artifact files.

Note that in our publish workflow here, we should make the github release depend on the prefix release, so that we are sure that's happened before we trigger this logic.

Tagging versions of the workflow releases repos is a separate but important / related concern.

cisaacstern:
closes #465


cisaacstern:
closing to re-do with better commit history

Issue #464, opened 2024-11-21
Better leveraging of spec.sha256, file headers, and release versions to ensure data consistency


cisaacstern:
Metaissue to track sub-issues chipped off of #430 for being a bit too difficult for the moment.

cisaacstern:
see https://github.com/wildlife-dynamics/ecoscope-workflows/pull/468#issuecomment-2490642506 for note on requirements section

cisaacstern:
Further notes on this:
- [ ] Can we get the actual version of ecoscope workflows (including scm version for dev) in the header?
- [ ] Also need to include ext-ecoscope version (that is, all dependencies from `requirements: `) ... but we can narrow down the version from the version specified in `requirements: `, which may be open-ended (e.g. >, *, etc.)
- [ ] Then we can made spec-sha256 -> spec-workflow-sha256. This can exclude `requirements: ` and any other `properties` that don't come in explicitly from the spec. We _do_ need some parsing in this, though, to account for variability in the way spec fields may be represented in YAML; e.g. the dasel dump[ is different](https://github.com/ecoscope-platform-workflows-releases/subject-tracking/pull/23/files) than the way we write by hand here in `examples`

Issue #320, opened 2024-10-16
publish-to-prefix is broken and fails silently


cisaacstern:
xref https://github.com/wildlife-dynamics/ecoscope-workflows/actions/runs/11358887750/job/31594157903#step:6:16

I can workaround for now by pushing these releases manually

cisaacstern:
This was an expired key, which I fixed, but leaving this open so we can address the silent failure aspect

atmorling:
This happened [again](https://github.com/wildlife-dynamics/ecoscope-workflows/actions/runs/11908373127/job/33183707442) 
I've updated to a key with a slightly longer shelf life

PR #468, opened 2024-11-21
Push examples workflows to releases repos, on release

### LINKED ISSUE #465: On release, push changes to releases directories, and auto-recompile there

cisaacstern:
closes #465

second take on #467 with fixed commit history


cisaacstern:
Reviewer notes:

- Here's an example of what this does which seems to check out https://github.com/ecoscope-platform-workflows-releases/events/pull/26
- I was able to push that from this PR by making the trigger for the workflow `on: push` just to make sure it worked. But I've now switched it back to `on: release` which is the behavior we want.
- In the `publish.yml` workflows, I changed the github release (which will trigger this workflow) dependent on the prefix release, because by the time we trigger the workflow added by this PR, we want to be sure the prefix release is up. (Note to self we should make `publish-to-prefix` fail hard if the token expires there, to prevent possible issues on this account; xref https://github.com/wildlife-dynamics/ecoscope-workflows/issues/320)
- Note that, as evidenced by https://github.com/ecoscope-platform-workflows-releases/events/pull/26, the actual diffs of the `spec.yaml` will be a bit different than the way we have them here in the examples, because https://daseldocs.tomwright.me/, which we are using for updating those files, doesn't roundtrip comments, and also seems to have its own opinions about how strings and arrays should be formatted in YAML, etc., which are not customizable AFAICT.
- This makes eyeballing the diffs between spec.yamls more-or-less impossible to do as a manual QA step. That's okay though. The ground truth for reviewing the bot PRs should probably not be based on visual similarity but rather: does the spec-sha256 in the file header on the compiled artifacts match those in the ecoscope-workflows? i.e., 
    
    https://github.com/wildlife-dynamics/ecoscope-workflows/blob/f9d5086e5e04d6555f68b49395e743fef6dd5f6f/examples/events/ecoscope-workflows-events-workflow/ecoscope_workflows_events_workflow/app.py#L3

- In theory, these hashes will be the same (or _could_ be the same 🤔 ) between `examples/` here in ecoscope-workflows, and in the compiled result, since the hash is based on the loaded YAML, not the raw text. As I write this, though, I realize that will not be the case because we are changing the `requirements: ` block, which will effect the hash. This suggests to me we should, as part of #464, remove `requirements: ` from this hash computation. I can do that tomorrow.

.github/workflows/push-examples-to-releases.yml
```diff
@@ -0,0 +1,33 @@
...
+      example: events
+      release_tag: ${{ github.event.release.tag_name }}
+      invoking_workflow_run: ${{ github.run_id }}
+    secrets:
+      GH_PAT: ${{ secrets.ECOSCOPE_ELEBOT_PAT }}
+
+  push-to-subject-tracking:
+    uses: ./.github/workflows/push-examples-reusable.yml
+    with:
+      example: events
```
cisaacstern, line 29:
```suggestion
  push-to-patrols:
    uses: ./.github/workflows/push-examples-reusable.yml
    with:
      example: patrols
      release_tag: ${{ github.event.release.tag_name }}
      invoking_workflow_run: ${{ github.run_id }}
    secrets:
      GH_PAT: ${{ secrets.ECOSCOPE_ELEBOT_PAT }}

  push-to-subject-tracking:
    uses: ./.github/workflows/push-examples-reusable.yml
    with:
      example: subject-tracking
```

PR #478, opened 2024-11-22
Fix hatch vcs versioning, overhaul fingerprint hashing, use fingerprint to ensure examples are not stale

### LINKED ISSUE #428: Make Spec.sha256 fingerprint evolve more slowly (and accurately)
Diffs like this

https://github.com/wildlife-dynamics/ecoscope-workflows/pull/427/files

are too big because of such `Spec` `@properties` include compilation variables such as jupytext parameters. These are needed for jinja, but are purely derived from the spec and do need to be included in "fingerprinting".

### LINKED ISSUE #433: Github Workflow to ensure all examples are recompiled before PR merge
All compiled files include a header comment which is (with the comments dropped) deliberately designed to be parsable TOML, e.g.

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/3887263264e16809337b23e5d0ba11f72f8c23af/examples/patrols/ecoscope-workflows-patrols-workflow/pixi.toml#L1-L3

This can tell us if the file is, in fact, compiled from the latest spec.yaml.

Let's actually start checking that to ensure we are correctly compiled before merging. Forgetting to recompile before merge _has_ happened to me 😄 .

### LINKED ISSUE #477: Fix hatch vcs versioning for better fingerprinting in generated file headers
We currently just see the fallback of `9999` which is problematic.

cisaacstern:
closes #428 
closes #477
closes #433

The hatch vcs versioning was not working due to a simple issue in the pyproject.toml.

Key points for this PR:
- The header comments on each file are now static. This is to prevent having massive numbers of files changed when the spec256 changes.
- I've moved the dynamic data from the headers (requirements versions and spec sha256, as well as a newly added params sha256) into a yaml "fingerprint" block in the README.
- In order to get the version of the requirements to reflect correctly in this fingerprint, I needed to change the compile environment from the editable install used for pytest, to a new dedicated `compile` env in the top level pixi.toml that actually uses the built releases of our source packages.
- ~~Re: the added params sha256, ultimately I would like this to be the basis for how we know whether or not a new release is a major or minor version bump (e.g. does it break existing config). But for now, it's more of a placeholder.~~ **Edit**: I ended up doing this as part of this PR after all.
- Requirements are now excluded from the spec.sha256, which should mean that we can use it as a sanity check to compare `examples/` hashes with released hashes (because the differing `requirements: ` blocks in each of those settings should no longer affect the hash).

Edit:

- As merged, this PR also now closes #433


cisaacstern:
I guess a unit test demonstrating that specs with different requirements have the same hash would be easy to add and helpful.

cisaacstern:
TODO:
- [ ] test to ensure the fix for #433 actually can fail

PR #480, opened 2024-11-22
Connection types

### LINKED ISSUE #479: handle different data connection types
Off the back of https://github.com/wildlife-dynamics/ecoscope-server/pull/189
Realised we need a lift here to enable RJSF parsing on the server side

atmorling:
closes #479


src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/io/_set_connection.py
```diff
@@ -6,10 +6,30 @@
...
+
+
+@task
+def set_gee_connection(
+    EcoscopeWorkflowsNamedConnection_GoogleEarthEngine: Annotated[
+        str,
+        Field(description="A named connection.", title="Name"),
+    ],
+) -> str:
+    return EcoscopeWorkflowsNamedConnection_GoogleEarthEngine
```
cisaacstern, line 35: Cool. Generally I am good with this and don't want to create unnecessary blockers but I did have one thought I'd like to discuss if you have a moment.

Apart from this smart / gee stuff, I was wondering if I made a bit of a strategic misstep in how these values are being substituted / parsed on the backend.

Rather than matching against parameter name, I wonder if should be matching against some type (basically, a string alias), which would make the rjsf post-processing be "just" a matter of replacing a `$def`?

examples/events/test-cases.yaml
```diff
@@ -8,7 +8,8 @@ test1:
       description: "A dashboard for visualizing events."
 
     er_client_name:
-      EcoscopeWorkflowsNamedConnection: "mep_dev"
+      EcoscopeWorkflowsNamedConnection:
+        name: "mep_dev"
```
cisaacstern, line 12:
```suggestion
      connection: "mep_dev"
```

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/io/_set_connection.py
```diff
@@ -1,15 +1,54 @@
...
+@task
+def set_smart_connection(
+    EcoscopeWorkflowsNamedConnection: Annotated[
+        SMARTConnection,
+        Field(description="A named connection.", title="Name"),
+    ],
+) -> SMARTConnection:
+    return EcoscopeWorkflowsNamedConnection
+
+
```
cisaacstern, line 46:
```suggestion
@task
def set_er_connection(
    connection: Annotated[
        EarthRangerConnection,
        Field(description="A named connection.", title="Name"),
    ],
) -> EarthRangerConnection:
    return EcoscopeWorkflowsNamedConnection


@task
def set_smart_connection(
    connection: Annotated[
        SMARTConnection,
        Field(description="A named connection.", title="Name"),
    ],
) -> SMARTConnection:
    return EcoscopeWorkflowsNamedConnection


```

I think we should consider adapting this pattern to allow for passing this value as a string, rather than an object

https://github.com/wildlife-dynamics/ecoscope-workflows/blob/8f4246440f74adf5da85d1afe44450a28adb9f20/src/ecoscope-workflows-ext-ecoscope/ecoscope_workflows_ext_ecoscope/connections.py#L95-L101

cisaacstern (commented):
I think we should consider adopting the pattern already used for the clients to resolve the connection from a string, rather than an object. This to me seems more readable from the RJSF layer (less nested), and also the python layer (since we can use a "regular" snake case parameter name).

Then the enum substitution on the server side can be changed to be based on the `$def` of these types (rather than the recursive search for the parameter names). This is more scalable and stable (i.e. `$def`-based find-and-replace, rather than reserved parameter name find-and-replace) as we look at other types of dynamic enum substitution going forward.

Finally, we will also need to change the parsing for the connection clients to be aware that they will be passed an EarthRangerConnection (and not a string).

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/io/_set_connection.py
atmorling, line 1: I'm wondering, since these defs are now specific to ER/SMART/GEE, does it make sense to move them into ext-ecoscope?

cisaacstern (approved):
LGTM!

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/io/_set_connection.py
atmorling: I think structurally (in my head and not necessarily in reality) I've thought of workflows core being the 'barebones'  workflows development infra, and ext-ecoscope being more implementation specific.

Based on your last point though it definitely makes sense to leave them where the are because based on what we have currently these aren't specific any one implementation.

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/io/_set_connection.py
```diff
@@ -1,15 +1,60 @@
...
+@task
+def set_smart_connection(
+    connection: Annotated[
+        SMARTConnection,
+        Field(description="A named connection."),
+    ],
+) -> str:
+    return connection.name
```
cisaacstern, line 50:
```suggestion
ConnectionName = Annotated[str, Field(title="Connection Name")]
DataSourceField = Field(
    title="Data Source",
    description="Select one of your configured data sources by name.",
)


@dataclass(frozen=True)
class EarthRangerConnection:
    name: ConnectionName
    connection_type: Annotated[
        Literal["EarthRanger"], AdvancedField(default="EarthRanger")
    ] = "EarthRanger"


@dataclass(frozen=True)
class SMARTConnection:
    name: ConnectionName
    connection_type: Annotated[Literal["SMART"], AdvancedField(default="SMART")] = (
        "SMART"
    )


@dataclass(frozen=True)
class GoogleEarthEngineConnection:
    name: ConnectionName 
    connection_type: Annotated[
        Literal["GoogleEarthEngine"], AdvancedField(default="GoogleEarthEngine")
    ] = "GoogleEarthEngine"


@task
def set_er_connection(
    connection: Annotated[
        EarthRangerConnection,
        DataSourceField,
    ],
) -> str:
    return connection.name


@task
def set_smart_connection(
    connection: Annotated[
        SMARTConnection,
        DataSourceField,
    ],
) -> str:
    return connection.name
```

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/io/_set_connection.py
```diff
@@ -1,15 +1,60 @@
...
 
 @task
-def set_connection(
-    EcoscopeWorkflowsNamedConnection: Annotated[
-        str,
-        Field(description="A named connection.", title="Name"),
+def set_gee_connection(
+    connection: Annotated[
+        GoogleEarthEngineConnection,
+        Field(description="A named connection."),
```
cisaacstern, line 57:
```suggestion
        DataSourceField,
```

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/io/_set_connection.py
cisaacstern: yes i agree it does sort of the challenge that model, which is also the one i'd had in my head as well

examples/patrols/spec.yaml
```diff
@@ -13,7 +13,7 @@ workflow:
 
   - name: Select EarthRanger Connection
```
cisaacstern, line 14:
```suggestion
  - name: Select EarthRanger Data Source
```

examples/events/spec.yaml
```diff
@@ -14,7 +14,7 @@ workflow:
 
   - name: Select EarthRanger Connection
```
cisaacstern, line 15:
```suggestion
  - name: Select EarthRanger Data Source
```

examples/subject-tracking/spec.yaml
```diff
@@ -13,7 +13,7 @@ workflow:
 
   - name: Select EarthRanger Connection
```
cisaacstern, line 14:
```suggestion
  - name: Select EarthRanger Data Source
```

cisaacstern (commented):
Combined with the fix mentioned in https://github.com/wildlife-dynamics/ecoscope-server/pull/189/files/5e09f15614911c040463587cff1c25758a1f8500#r1860992809 re: preserving styling information, I _think_ the changes below (or a variant of them) would let us have


Screenshot 2024-11-27 at 9 04 41 AM

src/ecoscope-workflows-core/ecoscope_workflows_core/tasks/io/_set_connection.py
atmorling: I think we keep an eye on stuff like this because maybe in the future we want to have something like `ecoscope-workflows-ext-common`

cisaacstern (approved):
tysm, alex!

Issue #514, opened 2024-12-03
Workflows pixi.toml's are imcompatible with pixi 0.39.0


atmorling:
See [here](https://github.com/wildlife-dynamics/ecoscope-workflows/actions/runs/12130520362/job/33821022484) for an example

It's unclear if this is an intended change from pixi or not, however this diff works for the parent pxi.toml:
```diff
- [feature.compile.dependencies.ecoscope-workflows-core]
- version = "*"
- channel = "file:///tmp/ecoscope-workflows/release/artifacts/"

- [feature.compile.dependencies.ecoscope-workflows-ext-ecoscope]
- version = "*"
- channel = "file:///tmp/ecoscope-workflows/release/artifacts/"

+ [feature.compile.dependencies]
+ ecoscope-workflows-core = {version = "*", channel = "file:///tmp/ecoscope-workflows/release/artifacts/"}
+ ecoscope-workflows-ext-ecoscope = {version = "*", channel = "file:///tmp/ecoscope-workflows/release/artifacts/"}
```

Also noting that in compiler.py we have:
https://github.com/wildlife-dynamics/ecoscope-workflows/blob/49736797f771a27fbe683291624db6818e4fda00/src/ecoscope-workflows-core/ecoscope_workflows_core/compiler.py#L825-L832

But the generated toml file looks like this:
https://github.com/wildlife-dynamics/ecoscope-workflows/blob/49736797f771a27fbe683291624db6818e4fda00/examples/events/ecoscope-workflows-events-workflow/pixi.toml#L19-L45

atmorling:
I've experimented with getting tomli_w to spit out the 'expected' structure here - seems like it's actually not possible based on stylistic choices it's making.

Might need to pin our CI to pixi==0.38.0 in the meantime.

cisaacstern:
I've opened https://github.com/prefix-dev/pixi/issues/2640 to track upstream

PR #516, opened 2024-12-03
explicitly install pixi v0.38.0

### LINKED ISSUE #515: Temporarily pin pixi to 0.38.0
Throwing this up as an interim option while #514 is addressed

atmorling:
Closes #515 
Xref #514 

Throwing this up as an option to unblock CI for the short term.
@cisaacstern feel free to reject this if there's a quick path to resolving #514 that I haven't thought of.


cisaacstern (commented):
Thanks so much for opening this and the related issue, both are very helpful to illustrate the failure mode.

I've found a fix that doesn't require pinning pixi in #517

cisaacstern:
superseded by #517

PR #408, opened 2024-11-12
Snapshot testing for examples

cisaacstern:
Motivated by regressions seen in the iframe outputs.

PR #521, opened 2024-12-04
Snapshot testing take 2 

cisaacstern:
supersedes #408


cisaacstern:
Remaining test failures are either intermittent failures (e.g. docker) or smart patrols failures as I have not yet pushed snapshots for smart pending confirmation that the images are sharable here.

cisaacstern:
Ah, I now see that subject tracking iframe comparison is in fact failing "legitimately", i.e. there is < 85 % (specifically, only ~77 %) similarity for at least one of the subject tracking iframes:

> https://github.com/wildlife-dynamics/ecoscope-workflows/actions/runs/12195806378/job/34022312505?pr=521

This will require some further investigation, as I expect we should be able to do better than this.

It strikes me that the subject tracking snapshots appear to be the least well-zoomed (that is, they are zoomed too far out), and so the actual map features are quite small and artifacted. It's possible this is the issue, and also that there may be some chromium launch args we could use to reduce variation.

cisaacstern:
A few other thoughts:
- If we push CLI testing through this approach as well, we can drop assertions from test-cases.yaml

PR #524, opened 2024-12-05
Fix dashboard filter selection

### LINKED ISSUE #523: #507 broke dashboard filter selection
https://github.com/wildlife-dynamics/ecoscope-workflows/pull/507/files

In #507 the property name of the filters was changed from `.index_name` to `.display_name`.

This broke filter selections because the views keys use `.index_name`.

I suggest we revert the property name back to `.index_name`, and we achieve the same visual effect by setting `"title":` to `.display_name`.

In my manual testing, the `ui:` fields we are generating currently don't seem to do anything, but putting a `"title":` field at the top level of the property object (alongside `"oneOf":`) seems to work on both our rjsf playground and the official playground.

atmorling:
Closes #523


cisaacstern (approved):
LGTM, thanks Alex.

Once it's done I am hoping that #521 will make these types of changes more visible (because it will snapshot the dashboard json as well as the iframes) but don't want to block this waiting for that to finish, so will put this in now.

